PR #299, opened 2014-11-30
Make requests lazy

rkneufeld:
Afternoon @ohpauleez!

I've largely finished implementing lazy request objects, and I wanted to solicit your feedback. `LazyRequest` now exists and it's functionality largely mirrors that of traditional maps. I have switched `request-map` generation in `servlet-interceptor` to use this new data structure.

## Differences from regular maps:

* `LazyRequest`s *are not* equal to the equivalent regular maps (even if the underlying map is identical). It turns out to be quite tricky (if not impossible) to get commutative equality, and even if you accomplish that, the semantics of equality are open for interpretation.
* To that end, `LazyRequest` implements `RawAccess` and `LazyDatastructure` protocols, which provide `raw`, `touch`, and `realized` methods on `LazyRequest`s.
  * `raw` returns the underlying map of an advanced data structure.
  * `touch` forces all delays in the underlying data structure to be eagerly realized.
  * `realized` returns a fully-realized, plain map.
* Iterating over a `LazyRequest` yields a sequence of `DerefingMapEntry`s (implementing `IMapEntry`). Just like `LazyRequest`'s `get`, a derefing map entry will defer realization of a key-value pair's value until it is explicitly accessed.
* `DerefingMapEntry` is not as close a match to `MapEntry` as `LazyRequest` is to `IPersistentHashMap`. `MapEntry` inherets a good deal of functionality from `AMapEntry`, which did not seem worth the effort unless it is absolutely necessary.

## Changes to `servlet-interceptor`

For the most part, `lazy-request` integrates quite cleanly to `io.pedestal.http.impl.servlet-interceptor`. The most notable omission, is that I have (for the moment) removed the transient/persistent bracketing around `request-map`'s more complicated `add-*` invocations.

Rather than jumping straight to complicating `LazyRequest` by implementing transient/persistent conversions, I wanted to check with you on the necessity of this optimization.

**Do we have any harnesses for measuring the performance gain/loss here? Similarly, it would be nice to benchmark `LazyRequest`s effect on existing applications.**

### DIFF
--- a/service/bench/io/pedestal/niotooling/ahttp.clj
+++ b/service/bench/io/pedestal/niotooling/ahttp.clj
@@ -1,0 +1,127 @@
+(ns io.pedestal.niotooling.ahttp
+;ns clj-ahttp.core
+  (:refer-clojure :exclude (get))
+  (:require [clojure.string :as cstr]
+            [clojure.core.async :as async])
+  (:import (com.ning.http.client AsyncHttpClient
+                                 AsyncHttpClientConfig
+                                 AsyncHttpClientConfig$Builder
+                                 AsyncHandler
+                                 AsyncHandler$STATE
+                                 RequestBuilder
+                                 HttpResponseHeaders)
+           (java.nio ByteBuffer)
+           (java.nio.channels Pipe
+                              ReadableByteChannel)))
+
+(defn request-method->str [request-method]
+  (-> request-method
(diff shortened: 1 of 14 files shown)


ohpauleez:
This code was hand-merged and updated on master.  Thanks for all the work!

PR #306, opened 2015-01-10
Add container functions to war deployment guide.

henriklundahl:


rkneufeld:
I'm not sure what to think of this. On one hand, it is certainly useful to include this somewhere, on the other, we've been working to reduce/eliminate the template in the long term.

Any thoughts on refreshing the guide over updating the template?

henriklundahl:
I didn't know you were working to reduce/eliminate the template. I think that removing it completely would make it harder to get started. The container functions aren't needed to get started, though.

Are you suggesting adding the functions in [service-war-deployment.md](https://github.com/pedestal/pedestal/blob/master/guides/documentation/service-war-deployment.md) instead? I guess that would be fine.

By the way, why is [ClojureVarServlet](https://github.com/pedestal/pedestal/blob/master/service/java/io/pedestal/servlet/ClojureVarServlet.java) including itself as first argument when calling the functions? When would one need access to it?

rkneufeld:
On Sat, Jan 10, 2015, at 04:16 PM, Henrik Lundahl wrote:
> I
 didn't know you were working to reduce/eliminate the template. I think
 that removing it completely would make it harder to get started. The
 container functions aren't needed to get started, though.


Yeah, at the moment it'd be too soon to drop it. We've got some work to
do around interfaces and documentation to get to the point I'd be
comfortable throwing people at it without the template.
> Are you suggesting adding the functions in
> service-war-deployment.md[1] instead? I guess that would be fine.


Exactly.
> By the way, why is ClojureVarServlet[2] including itself as first
> argument when calling the functions? When would one need access to it?


Not sure, I can only presume this is a byproduct of inheriting from
GenericServlet (from javax.servlet). **

-Ryan






Links:

  1. https://github.com/pedestal/pedestal/blob/master/guides/documentation/service-war-deployment.md
  2. https://github.com/pedestal/pedestal/blob/master/service/java/io/pedestal/servlet/ClojureVarServlet.java

ohpauleez:
Yeah, we should add these back into the template, commented out, with instructions for their use.

ohpauleez:
Added in: 002838599b7dc0eba33fbba55cb80f89fffb2c08

PR #307, opened 2015-01-26
reuse existing-token in assoc-double-submit-cookie

stuarth:
trivial change in `assoc-double-submit-cookie` to reuse `existing-token`


rkneufeld:
Looks good!

Issue #316, opened 2015-02-06
The io.pedestal.test/response-for function returns an empty response for routes with asynchronous interceptors.


briprowe:
When testing routes which have intercpetors that return a `core.async/channel` rather than a `context`, the `execute` function returns. This ultimately causes the the `test-servlet` to return before the `test-servlet-response` has been "realized".

By changing the `committed` atom to a promise, we can delay the thread running the test code until the response has been realized.

ohpauleez:
Thanks for using Pedestal and brainstorming ways we can improve it!  The more ideas, the better!

Rather than a promise, I'd like to see response-for use a channel internally.  This allows Pedestal to handle all async response capabilities, and removes the need for hard-coding timeouts (which should be avoided in async programming).  The final design isn't clear to me (I need to sit with it a bit).

jonahbenton:
Hey Paul, it looks like servlet-response-for needs a way to allow off-thread operations to complete  [in between the .service call and the retrieval of status and headers](https://github.com/pedestal/pedestal/blob/311646912a013cff1762efa9dd0deec0b5085777/service/src/io/pedestal/test.clj#L217). 

Is the thought to do a blocking take there from a channel set up in test-servlet-response, perhaps attached to response metadata?

ohpauleez:
Hi Jonah - that's the general idea.  More likely, we'll return the channel (that will receive the response-map), to allow developers to control the block characteristics (timeout and such).

candera:
I prefer to avoid breaking changes when possible. IMO, changing the return types of `response-for` (or anything, generally) is a breaking change. I don't think it's necessary in any event: I suspect we could introduce a new function - call it `response-for-async` - that would always return a channel. For a channel-returning interceptor, this would be the channel returned by the interceptor. For context-returning interceptors, this would be a channel that yields the context. `response-for` could be implemented on top of `response-for-async` by doing a blocking take on the result of `response-for-async`. This should ensure that nothing breaks while still letting people write tests against channel-returning interceptors without giving up control of scheduling - they would choose the semantics they want based on the function they used, and `response-for-async` is primitive enough to support all usages. That does assume people have a way to influence delivery of the value on the channel returned by the interceptor, but I'm not sure that's avoidable in any event.

Would something like that address what people need to do? I just started working on this, so it's entirely possible I'm missing something - would love to hear the use cases.

ohpauleez:
That all sounds accurate to me!  Thanks Craig!

[...]

PR #317, opened 2015-02-06
response-for should wait for the resp to be realized

briprowe:
This is a potential resolution for #316.

There are some magic numbers in this commit that hard-code how long to wait for the response. I think it would be good to make these parameters that the testing code can provide. I didn't add that to this commit, because that would change the api. Let me know if you think we should do that.

### DIFF
--- a/service/src/io/pedestal/test.clj
+++ b/service/src/io/pedestal/test.clj
@@ -143,7 +143,7 @@
                  (setContentType [this content-type] (swap! headers-map assoc :content-type content-type))
                  (setContentLength [this content-length] (swap! headers-map assoc :content-length content-length))
                  (flushBuffer [this] (deliver committed true))
-                 (isCommitted [this] (deref committed 10 false))
+                 (isCommitted [this] (realized? committed))
 
                  ;; Force all async NIO behaviors to be sync
                  container/WriteNIOByteBody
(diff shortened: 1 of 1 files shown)


ohpauleez:
Thanks for digging in and putting together this pull request!  I greatly appreciate it when people are willing to roll up their sleeves.  I'm going to close this particular PR - as stated in #316, I rather see a channel used than a promise with a timeout (for handling async responses).

briprowe:
You're right. Timeouts are a pain!

What does a more ideal solution look like? I presume it plays nicely with code doing SSE? Do you think the changes needed to get there might be (mostly) contained in `io.pedestal.test`? Or, do you think it may require deeper changes?

ohpauleez:
I'm not entirely sure what the ideal solution looks like.

I'd imagine the new `response-for` should work for all request formats (sync, async by channel, async by NIO, SSE, and soon WebSockets, which will be channel-based too).  For now, it'd be good enough just to work with sync and async behavior.  I think the changes are all contained within io.pedestal.test, and I think we can even keep backwards compatibility - allowing people to migrate up when needed.

briprowe:
How would this work with websockets?  Since a websocket is full-duplex, would we need 2 channels?

PR #324, opened 2015-02-27
send a 500 from error-debug

stuarth:


ohpauleez:
I've been thinking about this more - what's the use-case or advantage to doing this?  `error-debug` is used in development time only, and it'd be simple enough to add another interceptor in the stack that sets the status to whatever you want.

Can others weigh in on how they use `error-debug`?

stuarth:
Yep, def not a necessity, but it does make API client interactions nicer by default imo, e.g. a client expecting a json response will see the 500 instead of trying to parse the exception

ohpauleez:
After considering it more - I like the idea of the default scenario being 500 (exceptional).  Thanks!

PR #333, opened 2015-04-17
Bumped Immutant dep to 2.0.0 and rm'd obsolete sample

jcrossley3:
I'd rather not encourage folks with the 1.x sample.


ohpauleez:
Would you mind if I just updated the sample to illustrate the new `:immutant` option?

jcrossley3:
Not at all. I actually considered that, but decided it wouldn't look much different than your other samples, other than a couple lines. If you do go that way, would you mind using the name 'immutant' rather than 'jboss'? The latter is very overloaded and means different things to different people.

ohpauleez:
Will do!  Thanks Jim

PR #339, opened 2015-05-11
Fix docs: defhandler is in io.pedestal.interceptor.helpers namespace

Candid:


ohpauleez:
Hi Roman - thanks for the doc contribution!  I'm going to accept this for now, but going forward, we should really rewrite the docs to use the new interceptor API.

Candid:
Hey Paul, what is the new interceptor API?

ohpauleez:
The new [interceptor API](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/interceptor.clj) is in the original namespace.  It now is driven by a single function and a protocol (instead of a collection of many macros - which are now considered "helpers").  The change is documented in the [Pull Request](https://github.com/pedestal/pedestal/pull/301).  I now use the Interceptor API directly instead of the helpers for nearly everything.

You'll also notice that `Cons` and `List` are interpreted as the use-case for "interceptor-fn" - ie: You have a function that parametrizes and returns an interceptor.

Candid:
Ah, that makes sense. Thanks, Paul!

PR #343, opened 2015-07-09
add a couple of examples of creating interceptors

cap10morgan:
I had a hard time finding an example of the (slightly changed in 0.4.0) interceptor create syntax, so I thought I'd add it here.

Happy to move it elsewhere, improve it, etc.

### DIFF
--- a/guides/documentation/service-interceptors.md
+++ b/guides/documentation/service-interceptors.md
@@ -149,6 +149,39 @@ functions of further interceptors in the path are invoked as if no
 pause had occurred. A single context may pause and resume an arbitrary
 number of times.
 
+### Examples
+
+Here's how you define a "before" interceptor:
+
+```clojure
+(ns interceptor.example
+  (:require [io.pedestal.interceptor :refer [interceptor]]))
+
+(def my-before-interceptor
+  (interceptor
+   {:name ::hello-world
+    :enter
+    (fn [context]
+      (assoc context :response 
+                     {:status 200 :body "Hello world!" 
+                      :headers {"Content-Type" "text/plain"}}))}))
+```
+
+And an "after" interceptor:
+
+```clojure
+(ns interceptor.example
+  (:require [io.pedestal.interceptor :refer [interceptor]]))
+
+(def my-after-interceptor
+  (interceptor
+   {:name ::add-foo-header
+    :leave
+    (fn [context]
+      (update-in context [:response :headers] 
+                 merge "Foo" "Bar"))}))
+```
+
 ## Request Processing Across Threads
 
 This architecture allows for processing a single request across


guides/documentation/service-interceptors.md
```diff
@@ -149,6 +149,39 @@ functions of further interceptors in the path are invoked as if no
...
+  (:require [io.pedestal.interceptor :refer [interceptor]]))
+
+(def my-before-interceptor
+  (interceptor
+   {:name ::hello-world
+    :enter
+    (fn [context]
+      (assoc context :response 
+                     {:status 200 :body "Hello world!" 
+                      :headers {"Content-Type" "text/plain"}}))}))
```
jcf: Hi @cap10morgan,

We've used the built-in helpers a lot in our codebase for defining interceptors. I wonder if it'd be helpful to others if we mention these functions and macros?

``` clj
(require '[io.pedestal.interceptor.helpers :as interceptor])

(def my-before-interceptor
  (interceptor/before
    ::my-before-interceptor
    (fn [context]
      (assoc-in context [:request :example] true))))
```

Although I've avoided use of the macro forms to date, there are helpers that call `def` for you too. This snippet is copied out of a docstring in the helpers namespace:

``` clj
(defbefore parse-body-as-wibblefish
  [context]
  (let [request (:request context)
        new-request (assoc request :wibblefish-params
                           (wibblefish-parse (:body request)))]
    (assoc context :request new-request)))
```

https://github.com/pedestal/pedestal/blob/c6e4a1f7e9c411cc83305ddcc451a954ab931ddc/service/src/io/pedestal/interceptor/helpers.clj

ohpauleez: @jcf The macros are discouraged moving forward, and one should prefer using the Interceptor API directly where possible.

jcf: By "Interceptor API" does than mean `io.pedestal.http/interceptor` alone, or does that include the helper functions in `io.pedestal.interceptor.helpers`? We make pretty extensive use of the `before` function in our codebase, so I'd be sad to see it go.

ohpauleez: It means using `io.pedestal.interceptor/interceptor` directly - which has been overhauled.  It's simpler, better defined, and considerably less surface area.  The macros aren't going anywhere (they have a dedicated namespace), but they are discouraged and will break AOT compilation.

ohpauleez:
Thanks Wes!  Sorry it took me so long to get to these.  If you ever need help with anything Pedestal related, feel free to tweet or email me directly!

cap10morgan:
Cool, thanks @ohpauleez!

PR #348, opened 2015-07-22
Add a raw-response-for function to get at BAOS

tank157:
This allows us to get the raw bytes from a
response so we can, for example, test a PDF
being generated.


ohpauleez:
Thanks for contributing!  This change seems useful.  Have you signed a contributor's agreement?

tank157:
@ohpauleez I sent in our signed contributor's agreement today for the democracyworks organization.

ohpauleez:
Thanks @tank157 and thanks to Wes too!  If there's ever something that pops up in Pedestal, feel free to tweet or email me directly!

Issue #340, opened 2015-05-19
Exceptions for routes with path params


henriklundahl:
Pedestal 0.4.0 or 0.4.1-SNAPSHOT at 73b5854d.

Given this route:

    (defroutes routes [[["/a/:a/b/:b" {:get my-page}]]])

* I get a 500 response because of a NullPointerException [here](https://github.com/pedestal/pedestal/blob/73b5854d4ba01a6cd2146ca8a37ae6a2b76e8995/service/src/io/pedestal/http/route/prefix_tree.clj#L321) when requesting `/a/`, `/a/a` or `/a/a/b/`.
* When I request `/a`, `/a/a/b` or `/a/a/b/b/c` I get a 404 as expected.
* When I request `/a/a/` or `/a/a/b/b/` I get a 500 because of a StringIndexOutOfBoundsException [here](https://github.com/pedestal/pedestal/blob/73b5854d4ba01a6cd2146ca8a37ae6a2b76e8995/service/src/io/pedestal/http/route/prefix_tree.clj#L31).
* Requesting `/a/a/b/b` works fine.

ohpauleez:
Thanks for reporting this - My guess is that we're falling off the tree because of an off-by-one error when looking up the longest sequence.

snuggs:
404 is incorrect. The response code is actually 405 in this case. 404 is when the resource itself isn't defined. 
Given a resource collection definition @ `/foo/bar/` and a resource at `/foo/` a call to `/foo/bar` SHOULD respond with `405` status. 

Now of course `/sdafgagfgfg` SHOULD return `404` ;-)

PR #350, opened 2015-07-28
Add missing index.html used in benchmarks

jcf:
I'm working on a fix for #340 and came across a couple of small issues. This is the first, whereby when I start up a REPL, and when I try to reload namespaces I get an error because the web server benchmark namespace slurps a non-existent file when it's loaded.

My first inclination was to remove the side effect of slurping the file inside a function, but I don't know enough about this namespace to know if this will cause problems so I've gone with the more conservative solution of adding an `index.html` resource, and slurping that (I didn't see an obvious choice already in the repository). This shouldn't affect any benchmarks as the cost occurs when the namespace is loaded rather than at runtime, but please correct me if I'm missing something.

I've got a second PR incoming that enables some tests for the linear search router.

### DIFF
--- a/service/bench/io/pedestal/web_server_benchmark.clj
+++ b/service/bench/io/pedestal/web_server_benchmark.clj
@@ -1,10 +1,11 @@
 (ns io.pedestal.web-server-benchmark
-  (:require [io.pedestal.http :as http]
-            [io.pedestal.http.servlet :as servlet]
-            [io.pedestal.http.impl.servlet-interceptor :as servlet-interceptor]
-            [io.pedestal.interceptor.helpers :as interceptor :refer (defbefore)]
+  (:require [clojure.java.io :as io]
+            [io.pedestal.http :as http]
             [io.pedestal.http.immutant :as immutant]
-            [io.pedestal.http.jetty :as jetty])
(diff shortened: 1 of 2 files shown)


ohpauleez:
Thanks for reaching out and contributing to Pedestal!  That file already exists as a [test resource](https://github.com/pedestal/pedestal/blob/master/service/test/io/pedestal/public/index.html) - you could update your paths locally to include it.

Typically, developers aren't encouraged to develop directly against the benchmarking tooling, but instead some real-world application.  If you don't have an application to develop against, write a test to stand up a small app (follow the examples/tests that are already there), and that should be enough.
Also, feel free to keep using it if it's the easiest way for you to hack on Pedestal, I'd ask that you just modify your paths locally to make it work.

ohpauleez:
Also, note the comment at the top of the file - using this file (or even this benchmark) is discouraged in general.

jcf:
To clarify, I wasn't trying to use the benchmark file. It's on the classpath so gets reloaded whenever you refresh namespaces via `tools.namespace`. That means you get an error like this at the end of every refresh:

```
:error-while-loading io.pedestal.web-server-benchmark
#<CompilerException java.io.FileNotFoundException: ../index.html (No such file or directory), compiling:(web_server_benchmark.clj:66:58)>
```

PR #351, opened 2015-07-28
Enable trailing slash test for linear search router

jcf:
This is a small change that enables the `test-match-root-trailing-slash-map` test for the linear search router. Previously these tests were instead run twice with the prefix tree router.

I've signed the [Clojure contributor agreement](http://clojure.org/contributing) in the past. I'm not sure if I need to sign the Pedestal-specific contributor agreement too?


ohpauleez:
Thanks for contributing, but we're going to side step this still.  I definitely appreciate the effort though!

Ergotron-Mexico:
thanks.

PR #352, opened 2015-07-28
Trailing slash NPE

### LINKED ISSUE #340: Exceptions for routes with path params

jcf:
Given the following routes:

- `/people`
- `/people/:id`

…it was possible to get a 500 error via an NPE when visiting `/people/`, which obviously isn't what we want.

Based on an hour or so of reading lots of posts and opinion on whether URLs and RESTful routes should use trailing slashes I've concluded it's best to leave it up to the author (in this case the user of Pedestal). To support this I've made two changes to the prefix router. The first change checks a wildcard route not only contains a `/` but also that it's not the final character in the URL, and I've made sure we only ever call the `:payload` function for a route when it exists.

I can't say for sure this is the best solution, although I considered some alternatives like cleaning routes before inserting them, and enforcing certain good URL practices (forcing a trailing slash when adding a collection route) but opted for this approach because to my knowledge it will not break any existing Pedestal apps, and is relatively small/simple.

It's late here so please forgive any ambiguity. I'm happy to answer any questions you might have, and welcome feedback.

Fixes #340.

- http://googlewebmastercentral.blogspot.co.uk/2010/04/to-slash-or-not-to-slash.html
- http://programmers.stackexchange.com/questions/186959/trailing-slash-in-restful-api
- http://programmers.stackexchange.com/questions/186959/trailing-slash-in-restful-api
- http://sebastians-pamphlets.com/thou-must-not-steal-the-trailing-slash-from-my-urls/
- http://stackoverflow.com/questions/5948659/trailing-slash-in-urls-which-style-is-preferred
- http://www.bizcoder.com/the-mystery-of-the-trailing-slash-and-the-relative-url
- https://cdivilly.wordpress.com/2014/03/11/why-trailing-slashes-on-uris-are-important/
- https://github.com/nicolaiarocci/eve/issues/118
- https://www.safaribooksonline.com/library/view/rest-api-design/9781449317904/ch02s02.html (this one I particularly dislike as there's no evidence or justification for why this is the right approach)

---

**N.B.** This contains my small fix for enabling the trailing slash tests for the linear search router. I can remove that commit easily enough if necessary.

### DIFF
--- a/service/src/io/pedestal/http/route/prefix_tree.clj
+++ b/service/src/io/pedestal/http/route/prefix_tree.clj
@@ -264,7 +264,7 @@
 
         (:wild? node)
         (let [i (.indexOf path "/")]
-          (if (pos? i)
+          (if (and (pos? i) (not= (inc i) (count path)))
             (let [value (subs path 0 i)]
               (recur (assoc path-params (:param node) value)
                      (get-child node path (inc i))
@@ -303,6 +303,8 @@
                  (insert "/:foo/bar/:x" 4)
                  (insert "/:foo/baz/*rest" 5)))
 
+  (:path-params (lookup ptree "/foo/"))
+  ;;=> {:foo "foo/"}
   (:path-params (lookup ptree "/foo/bar"))
   ;;=> {:foo "foo"}
   (:path-params (lookup ptree "/foo/bar/baz"))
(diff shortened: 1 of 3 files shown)


ohpauleez:
Just a note - the error isn't with the linear router, it's with the prefix tree router.

jcf:
Thanks @ohpauleez.

jcf:
Hey @ohpauleez. I've got a passing build and a potential solution I'd appreciate your opinion on. I'm calling it a night, but will be around tomorrow if you come back with anything.

ohpauleez:
This issue was addressed in #361 - If you find the problem still pops up, open up a new issue.  Thanks!

PR #353, opened 2015-07-31
Fix failing interceptor tests.

ddeaguiar:
I was getting `java.lang.Exception: = expects more than one argument` exceptions when running the tests against `master`. This PR addresses that.


ohpauleez:
This should be fixed on master.  Thanks for spotting it!

PR #357, opened 2015-08-04
Remove excess function call

beorc:


ohpauleez:
Thanks for catching that.  Certainly a harmless thing, but no need to call it twice!

PR #361, opened 2015-09-01
Correctly 404 when no match.

### LINKED ISSUE #340: Exceptions for routes with path params

candera:
Added checks to handle nils that would NPE during route lookup,
especially in the face of trailing slashes.

Fixes #340. Probably addresses #352, too.

### DIFF
--- a/service/src/io/pedestal/http/route/prefix_tree.clj
+++ b/service/src/io/pedestal/http/route/prefix_tree.clj
@@ -28,7 +28,8 @@
   "Return the single character child key for the string started at
   index i."
   [s i]
-  (subs s i (inc i)))
+  (when (< i (count s))
+    (subs s i (inc i))))
 
 (defn- wild? [s]
   (contains? #{\: \*} (first s)))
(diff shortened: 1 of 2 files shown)


ohpauleez:
Fantastic, thank you!

PR #362, opened 2015-09-02
316 response for async

candera:
Make response-for work for interceptors that return channels. response-for still returns the whole body, by blocking until the returned channel closes.

### DIFF
--- a/service/src/io/pedestal/http/sse.clj
+++ b/service/src/io/pedestal/http/sse.clj
@@ -102,6 +102,30 @@
   ;;(end-fn)
   )
 
+;; This is extracted as a separate function mainly to support advanced
+;; users who want to rebind it during tests. Note to those that do so:
+;; the function is private to indicate that the contract may break in
+;; future revisions. Use at your own risk. If you find yourself using
+;; this to see what data is being put on `event-channel` consider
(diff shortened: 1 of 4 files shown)


candera:
@martintrojer After kicking it around with @ohpauleez for a bit, and trying out an approach (you can see the reverted commit in the branch), we decided to shelve `stream-for` for now. I'm of the opinion that testing SSE events at the level of the data put onto the event-channel is an application concern, not a Pedestal one. I did, however, pull out [this function](https://github.com/pedestal/pedestal/blob/316-response-for-async/service/src/io/pedestal/http/sse.clj#L105-L127) to support users with advanced needs. 

If you have any thoughts about something more clever we could do, we'd still love to hear them, but I wasn't able to figure out a good way to make this capability part of Pedestal at this time.

ohpauleez:
Thanks a ton @candera - great stuff as always!

martintrojer:
@ohpauleez @candera Hi, thanks for this work.

I did write some 'check-dynamic-var-channel-and-send-data-to-that-one-aswell' code in my `stream-ready-fn` for my tests while I was waiting for this work. It works fine, although somewhat ugly-fying the non test code.

I think it would be worthwhile writing an example on how to do these tests for SSE (and web sockets when they land).

PR #370, opened 2015-11-04
README to 0.4.1

stuarth:


ohpauleez:
You're faster than me!  :)  Thanks for always keeping up with the project.

PR #379, opened 2015-11-30
Changed 'content-length' type in the request header to 'string'

dalzony:
I know 'Pedestal incorporates existing ring middleware into its interceptors functionality.'
So, I've changed the type of the 'content-length' in the request header part to 'string', which satisfy the ring specifications.


ohpauleez:
Fixed on master

PR #387, opened 2016-01-04
Update csrf.clj

lambder:
Even after 'session token setting' the token will be present in :response not :request as assoc-session-token is working on response, therefore reading token from request would be null. 
Replaced to set the [:cookies "__anti-forgery-token"] explicit by parameter.


ohpauleez:
This is addressed on master now.  Note that to use the cookie-approach, the session tokens must first be processed (as per the comment in the source  Thanks so much!

PR #393, opened 2016-02-04
Update hello-world guide

dalzony:


dalzony:
Updated the generated example code (dependency versions).

dalzony:
:smile:

PR #394, opened 2016-02-05
Terse routing should provide friendly errors.

mtnygard:
* Add assertions to catch problems early while they are still related to
  the original data structures.
* Add some enforcement to catch nils that confuse verbose expansion
  later.
* Warn user if any of the input is left unused. (It probably indicates a
  problem.)

### DIFF
--- a/service/src/io/pedestal/http/route/definition.clj
+++ b/service/src/io/pedestal/http/route/definition.clj
@@ -13,9 +13,25 @@
 (ns io.pedestal.http.route.definition
   (:require [io.pedestal.http.route.definition.verbose :as verbose]
             [io.pedestal.interceptor :as interceptor]
+            [io.pedestal.log :as log]
             [clojure.set :as set]))
 
 (def schemes #{:http :https})
+
+(defn- unexpected-vector-in-route [spec]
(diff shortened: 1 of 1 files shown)


service/src/io/pedestal/http/route/definition.clj
```diff
@@ -31,10 +47,20 @@
...
+(defmethod expand-constraint :default [unmatched]
+  (assert false (unmatched-type-in-constraint unmatched)))
+
 (defprotocol ExpandableVerbAction
   (expand-verb-action [expandable-verb-action]
     "Expand `expandable-verb-action` into a verbose-form verb-map."))
 
+(def valid-handler?      (some-fn seq? symbol? interceptor/interceptor?))
+(def interceptor-vector? (every-pred vector? (comp :interceptors meta)))
+(def constraint-map?     (every-pred map?    (comp :contraints   meta)))
```
daemianmack: Typo: "contraints" => "constraints".

mtnygard: Good catch. I got bitten by recompiling the code that the macro uses, but not the namespace that uses the macro. Grr.

PR #395, opened 2016-02-17
Route table syntax

mtnygard:
Table routing should be documented.

* Moved old routing guide to reflect that it is specific to the terse syntax.
* Wrote a new routing guide for the table based syntax. The style is very different. The new guide starts with the simplest possible thing and builds up from there.
* Added a section on debugging routes, including reference to a new function `io.pedestal.http.route/try-routing-for` that lets the user try routes out from REPL without involving all the HTTP plumbing.

### DIFF
--- a/guides/documentation/service-routing.md
+++ b/guides/documentation/service-routing.md
@@ -253,7 +253,7 @@ You tell the router about constraints by supplying a map from
 parameter name to regular expression:
 
 ```clojure
-["/user/:user-id" :get view-user :constraints {:user-id #"[0-9]+"}
+["/user/:user-id" :get view-user :constraints {:user-id #"[0-9]+"}]
 ```
 
 If the constraint doesn't match, then the router keeps considering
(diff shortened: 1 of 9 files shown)


mtnygard:
This builds on PR #394.

guides/documentation/service-routing.md
```diff
@@ -12,601 +12,360 @@
...
 
 ```clojure
-["/user" {:get list-users
-          :post add-user}
- ["/:user-id"
- ^:constraints {:user-id #"[0-9]+"}
-  {:put update-user}
-  [^:constraints {:view #"long|short"}
-   {:get view-user}]]]
+["/user/:user-id" :get view-user :constraints {:user-id #"[0-9]+"}
```
kirillsalykin: Probably misses `]`.

mtnygard: Thank you, good catch!

ohpauleez:
Thanks so much!

PR #398, opened 2016-03-16
Revert "allow-origin arg as a map with :allowed-origins as seq of strings fix"

mtnygard:
After discussion with @ohpauleez, we think this change should be made in a different place. Moving it back for further review.


mtnygard:
Original PR #377

ohpauleez:
Thanks @mtnygard.  Let's see if this is just a bad doc, a violation of a lower-level code contract, or if the change should really be made in `normalize-args`.  Regardless, we should probably bring back those these as we move forward.

PR #403, opened 2016-03-16
Pedestal should have runtime metrics capabilities

ohpauleez:
This pull request enhances Pedestal with metrics capabilities.

It defines a metrics protocol (gauge, counter, histogram, meter) with a working backend bundled (dropwizard/codahale metrics) and a pass-through recorder (nil).  A public API is built around the `default-recorder`, which can be set via the service-map.

To setup a new recorder, add the key `:io.pedestal.http/metrics-init` - a 0-arity function that returns a new MetricsRecorder implementation.  Utility/Auxiliary functions are provided to help create tailored instances of a dropwizard MetricsRegistry, with tailored reporting.
The service-template is updated to have this key by default.

Metrics have been added for requests handled, open connections, open SSE streams, SSE payload sizes, and CORS


ohpauleez:
Don't merge just yet - I want to add a new sample that shows how to add StatsD and CloudWatch.

PR #405, opened 2016-03-17
HEAD requests should not return a body.

### LINKED ISSUE #372: HEAD Requests are not supported
It looks like HEAD requests are not supported at all in Pedestal.

There's the [head interceptor](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/ring_middlewares.clj#L85-L91) that [wraps the ring middleware for head requests](https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/middleware/head.clj), but it does it incorrectly.

The key thing to notice is the pedestal interceptor skips the wrap-head function, which keeps a reference to the original request that comes in. The consequence of not doing this is we simply assoc the :request-method to be :get to get the headers, then check on :leave if the changed :request-method is going to be :head, which of course it is not.

Assuming one writes this interceptor correctly, other issues crop up. It seems like the router does not route based on the request map, as we get a :servlet-request in the context that still says HEAD and our :get route never gets called. If this wasn't an issue, then by the default not found interceptor will return a 404 response because the ring middleware sets the response to nil.

Finally, if you try to explicitly create a :head route, that won't work because of [this vector](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/route/definition.clj#L183). I seem to be able to create custom, non-spec verbs and match :head requests by adding them in this vector.

And the explicit way isn't optimal because a :head request is simply the headers of the :get route. I'd like to be able to do it just once.

Are we, (@err, me) missing something simple?

mtnygard:
* This change removes Pedestal's use of the Ring head middleware.
* When the request method is :head, it will be replaced with :get during
  processing.
* During processing the context will also have a key :head-request? with
  value true.

Fixes #372

### DIFF
--- a/service/src/io/pedestal/http/ring_middlewares.clj
+++ b/service/src/io/pedestal/http/ring_middlewares.clj
@@ -88,7 +88,7 @@
   []
   (interceptor {:name ::head
                 :enter (fn [ctx]
-                         (if (= :head (-> ctx :request :request-method))
+                         (if (= :head (get-in ctx [:request :request-method]))
                            (-> ctx
                                (assoc :head-request? true)
                                (assoc-in [:request :request-method] :get))
--- a/service/src/io/pedestal/http/ring_middlewares.clj
+++ b/service/src/io/pedestal/http/ring_middlewares.clj
@@ -83,12 +83,20 @@
                 :leave (response-fn-adapter flash/flash-response)}))
 
 (defn head
-  "Interceptor for head ring middleware. If used with defroutes, it will not work
+  "Interceptor to handle head requests. If used with defroutes, it will not work
   if specified in an interceptors meta-key."
   []
   (interceptor {:name ::head
-                :enter #(update-in % [:request] head/head-request)
-                :leave (response-fn-adapter head/head-response)}))
+                :enter (fn [ctx]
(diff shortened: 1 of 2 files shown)


kirillsalykin:
Maybe I am a little to pedantic, but file named `ring_middlewares.clj` confuses me.

ohpauleez:
@mtnygard Small tweak - can you change `(-> ctx :request :request-method)` to `(get-in ctx [:request :request-method])

JakePi3:
Ideally the `:leave` function would also transform the request back to `:head` and drop the `:head-request?` key from the context.

The reason being that interceptors earlier in the chain which are about to have their `:leave` function called should see the request as a HEAD request, having no knowledge of `:head-request?` or any downstream transformations.

For example, if I wired in ring logger as an interceptor I'd get the following output... Notice the change in method when co-ordinating the start and finish.

```
2016-04-07 11:37:17.457+0100 INFO  default    ring.logger.tools-logging - Starting :head /about for 127.0.0.1 {"user-agent" "curl/7.30.0", "accept" "*/*", "host" "localhost:80"}
2016-04-07 11:37:17.474+0100 INFO  default    ring.logger.tools-logging - Finished :get /about for 127.0.0.1 in (18 ms) Status: 200
```

PR #417, opened 2016-03-29
Split SSE serialisation from dispatch, allowing interceptors to work with events

oliyh:
Hi,
I've made a change to split the SSE interceptor into two interceptors:
- `start-event-stream` - creates the channels, dispatch loop and response
- `serialise-event-stream` - serialises events on the SSE response body channel into the SSE format

The reason for the split is to allow other interceptors to collaborate on the events broadcast on an SSE stream before they are serialised (after which they are difficult to work with), in the same way the `json-body` interceptor allows other interceptors to work with Clojure data structures before serialising to a JSON string at the last moment.

My use case is for validating and serialising events in [pedestal-api](https://github.com/oliyh/pedestal-api) as described in [#5](https://github.com/oliyh/pedestal-api/issues/5).

To detect if the response is an SSE stream it currently stores and reads `::sse? true` on the context. I wasn't aware of a better way to work out which interceptor actually handled the request - something along the lines of how exceptions contain the throwing interceptor would have been useful, i.e. explicitly storing the name of the handling interceptor in the context. 

I've improved the SSE tests to check the events themselves to make sure this didn't break anything.

Let me know if you have any other suggestions and I'll gladly make adjustments.
Thanks!

### DIFF
--- a/service/src/io/pedestal/http.clj
+++ b/service/src/io/pedestal/http.clj
@@ -21,6 +21,7 @@
             [io.pedestal.http.servlet :as servlet]
             [io.pedestal.http.impl.servlet-interceptor :as servlet-interceptor]
             [io.pedestal.http.cors :as cors]
+            [io.pedestal.http.sse :as sse]
             [ring.util.mime-type :as ring-mime]
             [ring.util.response :as ring-response]
             [clojure.string :as string]
@@ -223,6 +224,7 @@
                      (or enable-session enable-csrf) (conj (middlewares/session (or enable-session {})))
                      enable-csrf (conj (csrf/anti-forgery enable-csrf))
                      true (conj (middlewares/content-type {:mime-types ext-mime-types}))
+                     true (conj sse/serialise-event-stream)
                      true (conj route/query-params)
                      true (conj (route/method-param method-param-name))
                      (not (nil? resource-path)) (conj (middlewares/resource resource-path))
@@ -321,4 +323,3 @@
 
 (defn servlet-service [service servlet-req servlet-resp]
   (.service ^javax.servlet.Servlet (::servlet service) servlet-req servlet-resp))
-
--- a/service/src/io/pedestal/http/sse.clj
+++ b/service/src/io/pedestal/http/sse.clj
@@ -54,27 +54,30 @@
 (def ^ThreadFactory daemon-thread-factory (counted-thread-factory "pedestal-sse-%d" true))
 (def ^ScheduledExecutorService scheduler (Executors/newScheduledThreadPool 1 daemon-thread-factory))
 
-(defn mk-data [name data]
-  (let [bab (ByteArrayBuilder.)]
-    (when name
-      (.write bab ^bytes EVENT_FIELD)
-      (.write bab ^bytes (get-bytes name))
-      (.write bab ^bytes CRLF))
+(defn mk-data
+  ([event] (mk-data (if (map? event) (str (:name event)) nil)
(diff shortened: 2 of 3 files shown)


ohpauleez:
Hi @oliyh!  Thanks for using Pedestal and seeking to make it better.  This is certainly an interesting idea.

Recently, work was done to enable interceptor chains to be run in one direction (that is, chains can be processed as in, out, in/out).  The idea is that one can express chains to process streams of data (for SSE, WebSockets, whatever), and integrate them into endpoint handling within their application/service.

In short, I also recognize the problem you're encountering.  Would additional interceptors at the SSE-endpoint level solve your specific validation problem?

oliyh:
Hi,

My aim for pedestal-api is to produce a set of interceptors that cover the vast majority of boilerplate required to make a swagger compliant API. One of these interceptors is the content negotiation which sits near the end of the leave phase, serialising clojure data structures into whatever the client requested.

The application developer would want to put a clojure data structure in to the SSE channel and let pedestal-api take care of serialising it. That means it's important that there is an opportunity to transform the data on the channel before it's turned into a byte array - which is what this PR provides.

The SSE interceptor could take more arguments (I.e. another interceptor chain to process events, which I think you're alluding to) but then it would be returning a channel which would pass through all the other interceptors upon which the data would already be serialised and therefore inscrutable. I think it would be better if it was data structures all the way until the last possible moment, in case they have an opinion on the response, which necessarily means the serialising interceptor is one of the very last in the entire chain.

Hope this makes sense, I could put together an example of how pedestal-api would use this PR, or similar, if it would help clarify.

Cheers

ohpauleez:
I certainly understand the core problem (There is no way to post-process data/events going into an SSE stream before they're serialized).  I don't like splitting SSE into two special-cased interceptors for a handful of reasons, but my main concerns are:
 - SSE events/streams don't fit the majority of processing done by base interceptor chain
 - Post-processing events is best described by a chain specific to that stream

I'm going to keep thinking about this.  Do you have any other examples of backend libraries that provide content negotiation for SSE, or real-world systems that utilize content-negotiation for SSE?

oliyh:
Hi,

I agree, SSE is a special case. I can understand that exiting the chain of HTTP interceptors and instead going through something SSE specific would be desirable, but it would also add complexity. A trade off, as always.

I've had a look for examples of content negotiation / content types for SSE, and can't find any. I'm not really sure why - possibly because the de facto standard is JSON and no one really thought that anything else would be needed (my use case here is edn/transit for things like date objects, sets, and keywords). I think it would be a nice feature to implement as part of pedestal-api in any case.

What if the SSE handler added the `serialise-event-stream` interceptor dynamically? That would actually be a better implementation than my use of the `:sse?` flag.

ohpauleez:
Hi @oliyh 

Interceptors that enqueue more interceptors are definitely mode idiomatic in Pedestal for core interceptors.  Still, I think specifying and interceptor chain for the SSE endpoint targets the root cause/core need better.

I'm still weighing the trade-offs.  I want to take some time and write a few applications in each style.

Thanks again for being so engaging, finding this issue, and proposing a solid change!

ohpauleez:
@oliyh Is this still a standing issue for you?  If so, let me know and we can work through the remaining pieces.

oliyh:
Hi @ohpauleez I will see if there is a way to do what I want to and let you know, thanks.

ohpauleez:
I'm closing out this PR since there hasn't been much movement on it.  Please re-open with new code based on latest Pedestal code/features.  Thanks so much!

Issue #384, opened 2015-12-27
SSE channel seems to close after a single event is sent


batasrki:
This is probably down to how I'm attempting to use the SSE feature rather than a bug. As the title says, I'm able to only receive one SSE message on the client side, which isn't what I expect. Unlike the samples code, I am actually trying to integrate the SSE into a background pipeline where I post a URL, its HTML gets fetched and `<title>` parsed, and db record updated. As a last step, I'd like to post the updated record on the SSE stream such that the client receives it and updates itself.

The example code and the log is in this gist, https://gist.github.com/batasrki/74e4e62974eb8dde3e84.

Any help is appreciated.

ohpauleez:
Hi @batasrki!  Thanks for using Pedestal and reaching out for help.  In the future, you might want to use the [mailing list](https://groups.google.com/forum/#!forum/pedestal-users) - more people are available there to help.

In the gist, the code is correctly only sending one event.  A `go` block, only runs once, just like a normal block.  You'll either need to use a `go-loop` or use `loop` within your go-block.

A few additional things you should be aware of:
 * The `stream-ready-fn` is usually to just setup the channel (potentially store it in an atom), not really for sending your events.
 * The `stream-ready-fn` takes [two args](https://github.com/pedestal/pedestal/blob/master/samples/server-sent-events/src/server_sent_events/service.clj#L40), The event channel (that you should be using) and the Pedestal context for that given request.
* "Broken Pipe" errors usually mean the other end (client side) has closed the connection, but Jetty wasn't aware of it (because of TCP keepalive).

If you fix your code up, and you're still getting errors, just re-open this bug. Thanks!

batasrki:
Apologies for having to re-open this issue again, but I made the suggested changes (I believe) and got the same result. I then posted to the mailing list, also as suggested, here, https://groups.google.com/forum/#!topic/pedestal-users/J8oj33cg_e0. There has been no response there and I'm at a loss.

Any help is still appreciated.

batasrki:
Also, I'm unable to re-open this issue.

jonpither:
I have the same issue

ohpauleez:
Let me update the sample code and see if I can recreate.  I haven't seen this behavior myself.

ohpauleez:
The SSE sample works correctly: https://github.com/pedestal/pedestal/tree/master/samples/server-sent-events

ohpauleez:
[...]

PR #420, opened 2016-04-11
Don't log complete stacktrace for disconnected clients

djjolicoeur:
For https://github.com/pedestal/pedestal/issues/384

@ohpauleez I know it's a pedantic point, but can we handle the known case a little more elegantly than logging an error with a stacktrace?  That error is constantly being logged in our production logs as people disconnect and refresh and is a source of confusion while trying to troubleshoot real issues.  Since the exception is being logged, rather than being thrown, theres not much we can do to filter it out unless I want to filter _all_ error logs from that namespace, which doesn't seem like a reasonable solution.  This PR catches anything that extends `EOFException`, logs a smaller message at the warn level, and closes the body.  Any other Throwable will still log the stacktrace for debugging and the behavior of the application remains unchanged.

### DIFF
--- a/service/src/io/pedestal/http/impl/servlet_interceptor.clj
+++ b/service/src/io/pedestal/http/impl/servlet_interceptor.clj
@@ -25,7 +25,7 @@
             [ring.util.response :as ring-response])
   (:import (javax.servlet Servlet ServletRequest ServletConfig)
            (javax.servlet.http HttpServletRequest HttpServletResponse)
-           (java.io OutputStreamWriter OutputStream)
+           (java.io OutputStreamWriter OutputStream EOFException)
            (java.nio.channels ReadableByteChannel)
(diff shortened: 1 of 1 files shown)


ohpauleez:
I'm totally ok with this change, but I'm going to make it lower in the stack (so we can catch it for all cases with the client bails out on an open connection).

Thanks for jumping in and putting the proposed change together though!

I'll leave this open until I land the lower change.

djjolicoeur:
@ohpauleez that would be great, thanks!

ohpauleez:
@djjolicoeur You made it at the correct spot, I just wanted to add the same handling for the sync handling (same error situation happens there when the client breaks the pipe).

The changes are sitting on the uber branch, [interceptor-providers](https://github.com/pedestal/pedestal/tree/interceptor-providers), which should land on master soonish.

PR #421, opened 2016-04-19
Interceptor chain overhaul, Servlet independence, Websockets, and Routes

ohpauleez:
This branch integrates a handful of changes across all of Pedestal.  As always, where possible, breaking changes were avoided (or patched into the new stuff).

### Interceptor chain dynamics

The interceptor chain is now a proper, public namespace ([io.pedestal.interceptor.chain](https://github.com/pedestal/pedestal/blob/interceptor-providers/service/src/io/pedestal/interceptor/chain.clj)).  In addition to the standard public API functions `execute` and `enqueue`, there's now `execute-only`.  Execute-only allows running a chain in a single direction.  This enables the use of standalone interceptor chains for processing any stream (internally within the application, SSE, WebSockets, etc.).  At some point in the future, the interceptor pieces will become a standalone module within the Pedestal project.  I encourage all to read through the [interceptor test](https://github.com/pedestal/pedestal/blob/interceptor-providers/service/test/io/pedestal/interceptor_test.clj)

### Interceptor chain providers

Until this branch, Pedestal has always needed a Servlet container.  The very base of the interceptor chain was bootstrapped directly against the Servlet Interface.  Interceptor Chain Providers remove that coupling by exposing another control to the developer.  Chain Providers bootstrap an interceptor chain to a lower-level container.  Pedestal no longer makes any assumptions about the lower-level container, whatever the Chain Provider exposes/handles, Pedestal will use.  Providers are specified in the service map.
The core changes to enable this are in place along with a [sample](https://github.com/pedestal/pedestal/tree/interceptor-providers/samples/chain-providers) illustrating how write your own providers.  More polish and refactoring will happen as we write more non-servlet platforms.

### Websocket support for Jetty

Along with ALPN and HTTP/2 support, Pedestal now ships with WebSocket support for Jetty.  Websockets are specified outside of route definitions (since they are container specific). See the [sample](https://github.com/pedestal/pedestal/tree/interceptor-providers/samples/jetty-web-sockets) for more details.

### Route definition refactor

The route specification and definitions within Pedestal were reorganized and refactored.  Routing formats are all programmed against a single protocol (ExpandableRoutes).  The `defroutes` macro is no longer needed and will be deprecated at some point - Pedestal consumes routing data structures that satisfy ExpandableRoutes directly.  This change also greatly reduces the "surface area" of route definitions in Pedestal.  The service template now illustrates routes in all three major formats (Vector, Map, and Set/Table).
This refactor also prepared the Routes and Router to be standalone modules in the future.  Some of these changes are visible on the [recent samples](https://github.com/pedestal/pedestal/blob/interceptor-providers/samples/chain-providers/src/chain_providers/service.clj)

### Quiet errors on client disconnect

Pedestal previously logged errors when the client disconnected and broke the pipe.  These EOFExceptions will now be WARN messages for synchronous and asynchronous responses.

### Logging and Runtime Metrics

Pedestal's logging was refactored against a protocol, which will fallback onto SLF4J.  You can short-circuit all logging by configuring the logger to `nil`.  You can now also override the logger per log message - Loggers must satisfy the LoggerSource protocol.
Pedestal also now ships with runtime metrics support - counters, meters, gauges, and histograms.  These metrics can be published to any source/recorder.  There is a [sample](https://github.com/pedestal/pedestal/tree/interceptor-providers/samples/helloworld-metrics) on how to publish metrics to StatsD.  The same approach can be used to publish the metrics to CloudWatch, SLF4J, or anything else you might have in production.  By default, metrics are published to JMX.

Internally Pedestal tracks active connections, reqs/s, errors/s, SSE payload histograms, and a few others.  Understanding an application rate is as easy as `(log/meter ::my-stat-tag`).  Take a look at the sample!


ohpauleez:
I'd love for people to comment on these changes, especially if you see something that will break your use of Pedestal in horrible ways.  Similarly, I want to see how much support these changes have before I land them.

Consider this PR a "Release Candidate" of sorts.

PR #423, opened 2016-04-23
Service should always keywordize keys in :params key of context

### LINKED ISSUE #240: Service should always keywordize keys in :params key of context
From #238, in a request context POST `:params` keys are strings, while GET keys are keywords.

After talking with @timewald we believe it is reasonable going forward for 0.3.x to *always* return keywords on the `:params` key of context.

**Acceptance Criteria:**

* [ ] `:params` is always keywordized. This can probably be accomplished by adding  [`middlewares/keyword-params`](https://github.com/pedestal/pedestal/blob/0719c9f77b12b25352a3c5a838273eedfa24798b/service/src/io/pedestal/service/http/ring_middlewares.clj#L89-L91) to [default-`interceptors`](https://github.com/pedestal/pedestal/blob/0719c9f77b12b25352a3c5a838273eedfa24798b/service/src/io/pedestal/service/http.clj#L107-L154)
* [ ] New tests are added to enforce this behavior. Old tests are updated.
* [ ] `:form-params` and `:query-params` behavior is not unilaterally changed.

mtnygard:
* Parameters are keywordized, whether from query string, form params, or
  body params.

This change became bigger than intended. In particular, we've been
leveraging the Ring middleware for keywordization. There are two
challenges there. First, the syntax doesn't support namespaced
keywords. Second, the key function is not public.

Taking a longer view, I think we need to reduce Pedestal's reliance on
Ring middleware, migrating instead to Pedestal-native interceptors for
equivalent functionality. This commit does that for the `keyword-params`
interceptor. The var `io.pedestal.http.ring-middlewares/keyword-params`
is now an alias to `io.pedestal.http.params/keyword-params`

Fixes #240

### DIFF
--- a/service/src/io/pedestal/http/body_params.clj
+++ b/service/src/io/pedestal/http/body_params.clj
@@ -15,6 +15,7 @@
             [cheshire.core :as json]
             [cheshire.parse :as parse]
             [clojure.string :as str]
+            [io.pedestal.http.params :as pedestal-params]
             [io.pedestal.interceptor.helpers :as interceptor]
             [io.pedestal.log :as log]
             [cognitect.transit :as transit]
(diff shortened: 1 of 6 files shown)


ohpauleez:
Excellent to see and thanks for the hard work!  I was *just* rewriting other ring middleware-wrappers this morning, so this is in-line with other local changes I have.

ohpauleez:
Thanks!

PR #433, opened 2016-05-25
Fixed the namespace and var name for table routes

christianromney:
The names in the documentation are out of date with the refactoring.


ohpauleez:
Thanks a ton!

PR #435, opened 2016-06-01
There is only one fact at the top of the interceptors documentation.

bayan:
Minor wording change in documentation for interceptors.


ohpauleez:
Thanks for the grammar check!  Interestingly enough, with the new interceptor API, I'm not so sure they're more complicated to write anymore :)

PR #440, opened 2016-06-06
CSRF should use keywordized form-params

terjesb:
Since 0.5.0, form-params now use keywords instead of string keys.


ohpauleez:
Thanks for spotting the issue and submitting the pull request!  I'm going to be in the CSRF code this week, so I'll pull in the changes.

ohpauleez:
This is fixed on master now.  Thanks!

PR #446, opened 2016-06-17
Avoid double-catch in go block bug when SSE client disconnects

oliyh:
Hi,

It looks like the recent change to quietly shutdown the async response loop when a client disconnects (SSE streams in particular) is falling into the trap laid by http://dev.clojure.org/jira/browse/ASYNC-169, namely there are two `catch` blocks inside a `go` block which sends everything into a tailspin.

On a deployed webserver you will see that async-macro-threads never get cleaned up, and eventually you will eat up all the async dispatchers and everything will lock up. This was made much more likely to happen by the decrease of the default pool size to 8 in the 0.2.385 release of core.async (this is when we came across it).

In the unit test, you will find that with the old source code it will lock up and hang your VM as described in ASYNC-169. With the fixed source code, you will find that it errors instantly as expected.

Hope this helps
Oliy

### DIFF
--- a/service/src/io/pedestal/http/impl/servlet_interceptor.clj
+++ b/service/src/io/pedestal/http/impl/servlet_interceptor.clj
@@ -104,17 +104,16 @@
           (try
             (write-body servlet-response body-part)
             (.flushBuffer ^HttpServletResponse servlet-response)
-            (catch EOFException e
-              (log/warn :msg "The pipe closed while async writing to the client; Client most likely disconnected."
-                        :exception e
-                        :src-chan body)
-              (async/close! body))
             (catch Throwable t
-              (log/meter ::async-write-errors)
-              (log/error :msg "An error occured when async writing to the client"
-                         :throwable t
-                         :src-chan body)
-              (async/close! body)))
+              (if (instance? EOFException t)
+                (log/warn :msg "The pipe closed while async writing to the client; Client most likely disconnected."
+                          :exception t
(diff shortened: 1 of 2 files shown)


ohpauleez:
Thanks for spotting this!

What is the rationale behind the `(finally ... async/close)`?  Why close the channel eagerly in the good case?  Previously the channel was only ever closed eagerly by the consumer (Pedestal) in the error case to prevent the consumer from continuing forward.  Otherwise it's the producer's responsibility to close the channel.

oliyh:
The old code had the `close!` in both catch blocks, so that was a small simplification which retained the same behaviour. But whether on error or on disconnect, wouldn't you want to close the channel anyway? In both cases you can't write to it anymore, even if one is part of normal lifecycle and the other is not.

oliyh:
Ah I have just re-read the code more carefully and yes, you're right, you don't want to close the body channel in the finally block. My mistake - it's been a long day staring at core.async internals and trying to track this one down! I agree with your version in master.

oliyh:
By the way thanks for getting this to master so quickly. Now that the major changes are done will Pedestal be moving to a faster release cadence for smaller fixes like this?

ohpauleez:
I think releases (especially point releases) will happen much quicker now that the planned changes are all done.  After 0.5.1, there's not much left to do before the official 1.0 release.  Thank you for all your help, input, and feedback along the way!  And also for building up the community tooling!

PR #447, opened 2016-06-27
Handle empty transit body params

r0man:
When sending a request with a `Content-Type` header of
`application/transit+json` or `application/transit+msgpack` and an empty
body, Pedestal throws a `java.io.EOFException` exception. This happens
because the Transit library doesn't read from an empty
ByteArrayInputStream. This does not happen with EDN, JSON or form
params. This patch fixes that problem by checking the available bytes of
the input stream and makes the behaviour consistent with the other
content types.

### DIFF
--- a/service/src/io/pedestal/http/body_params.clj
+++ b/service/src/io/pedestal/http/body_params.clj
@@ -122,10 +122,12 @@
   body of that request with `transit/read`. options is a sequence to
   pass to transit/reader along with the body of the request."
   [& options]
-  (fn [request]
-    (assoc request
-      :transit-params
-      (transit/read (apply transit/reader (:body request) options)))))
+  (fn [{:keys [body] :as request}]
+    (if (zero? (.available body))
+      request
+      (assoc request
+             :transit-params
+             (transit/read (apply transit/reader body options))))))
 
 (def transit-parser
   "Take a request and parse its body as transit."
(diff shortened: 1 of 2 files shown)


ohpauleez:
So funny that you hit this, @solussd and I hit this just last week as well.  Sorry about the issue, but thanks for reporting and thank you for submitting a patch!

ohpauleez:
@r0man Have you signed a Cognitect contributors agreement already?

r0man:
@ohpauleez Yes, I think I signed it 3 years ago. It that's still the one.

ohpauleez:
Still is the one, thanks!

r0man:
Cool, thanks also! :)

r0man:
@ohpauleez Are there any plans for a release with this patch? If not, could you please push a snapshot to clojars?

ohpauleez:
It will be pushed out in the next (0.5.1) release.  Not sure when that'll get wrapped up (should be soonish), but until I suggest you use a local install.  Generally speaking, the Pedestal project doesn't push snapshot releases.

r0man:
Yes, I'm using a local snapshot at the moment. I just thought snapshots make it easier for people to test stuff on Travis without having to tweak their build scripts. Anyway I'll find a solution. :)

PR #453, opened 2016-08-04
Refactor transit-body and accept Transit options

christianromney:
The `transit-json-body` and `transit-msgpack-body` interceptors shared the majority of their implementations. This commit refactors them to call a new `transit-body-interceptor`. This new function also allows the caller to supply optional transit-options which can be used to install custom Transit write handlers.

Signed-off-by: Christian Romney <cromney@pointslope.com>


ohpauleez:
Looks great, thanks so much!

PR #461, opened 2016-08-29
Add test for parsing Transit request bodies through response-for.

grzm:
`response-for` didn't implement `.available` on the ServletInputStream,
which caused it to start failing with the fix to handle empty body
params: d948b6122bbe086562419efa36446f9bd330115c

`response-for` was fixed in 4b5f792dbe81de4c6521aa1e3b82f46eea3e42f6,
and the added test covers this fix. While this latest fix is for
`response-for`, it also provides additional coverage for parsing
(non-empty) Transit bodies.

Change-Id: Ib1fa04f6f52119a9c32a1e2783994f2bacbe306c

### DIFF
--- a/service/test/io/pedestal/http_test.clj
+++ b/service/test/io/pedestal/http_test.clj
@@ -18,6 +18,7 @@
             [io.pedestal.interceptor.helpers :as interceptor :refer [defon-response defbefore defafter]]
             [io.pedestal.http.impl.servlet-interceptor :as servlet-interceptor]
             [io.pedestal.http.route.definition :refer [defroutes]]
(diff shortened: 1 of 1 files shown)


grzm:
I decided to add the test to `http-test` as response-for is also exercised here. I did need to bring in `body-params`, which isn't as clean as I'd like. `test-test` didn't seem like a good fit as it's testing more utility functions. I wonder if there's a place for more "integration", "functional", or "app" testing that cuts across the modules? For example, this tests something specifically for `response-for` (part of test) and `body-params` (part of http).

ohpauleez:
Looks good!

grzm:
Thanks!

Issue #459, opened 2016-08-29
(csrf/anti-forgery {:cookie-token true}) returns 404 if session isn't set


grzm:
# Description

When using `(csrf/anti-forgery {:cookie-token true})`, a request that does not include a token gets a 404 Not Found response rather than a 403 Forbidden.

## Expected Behavior

If the request does not include a valid CSRF token, the response should be 403 Forbidden.

## Actual Behavior

A request without a valid CSRF token receives a 404 Not Found response.

## Steps to reproduce

Test case repo: https://github.com/grzm/pedestal-csrf-anti-forgery-cookie-404

In brief:
```clojure
(defn home-page
  [request]
  (ring-resp/response "Hello World!\n"))

(def common-interceptors [middlewares/cookies
                          (middlewares/session {})
                          (body-params/body-params)
                          http/html-body])

(def csrf-interceptors [middlewares/cookies
                        (middlewares/session {})
                        (csrf/anti-forgery)
                        (body-params/body-params)
                        http/html-body])

(def csrf-cookie-interceptors [middlewares/cookies
                               (middlewares/session {})
                               (csrf/anti-forgery {:cookie-token true})
                               (body-params/body-params)
                               http/html-body])

(def routes #{["/csrf" :post (conj csrf-interceptors `home-page)
               :route-name :post-csrf]
              ["/csrf-cookie" :post (conj csrf-cookie-interceptors `home-page)
               :route-name :post-csrf-cookie]})
```

```
λ curl -v localhost:8080/csrf -X POST
*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> POST /csrf HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.46.0
> Accept: */*
> 
< HTTP/1.1 403 Forbidden
< Date: Mon, 29 Aug 2016 15:09:58 GMT
< Strict-Transport-Security: max-age=31536000; includeSubdomains
< X-Frame-Options: DENY
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block
< Content-Type: text/html
< Transfer-Encoding: chunked
< Server: Jetty(9.3.8.v20160314)
< 
* Connection #0 to host localhost left intact
<h1>Invalid anti-forgery token</h1>


λ curl -v localhost:8080/csrf-cookie -X POST
*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> POST /csrf-cookie HTTP/1.1
[...]

PR #462, opened 2016-08-29
Pass through response if missing token when using double-submit cookies.

grzm:
The missing token will already fail on entering the csrf/anti-forgery
interceptor, and there may be no token to find when leaving. Keep
the 403 Forbidden response intact rather than returning nil, which
will be interpreted as a 404.

Change-Id: I99f07735eedaefb6fd9e1d20946075e1395c7857

### DIFF
--- a/service/src/io/pedestal/http/csrf.clj
+++ b/service/src/io/pedestal/http/csrf.clj
@@ -48,10 +48,11 @@
 ;; This must be run after the session token setting
 (defn- assoc-double-submit-cookie [request response]
   ;; The token should also be in a cookie for JS (proper double submit)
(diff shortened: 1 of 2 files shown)


grzm:
I *think* I've interpreted what's going on correctly. Given this covers an area of security and I am *not* a security guru, I definitely want other eyes on this. In particular, in this case should we *force* a 403 rather than just pass through the existing response?

grzm:
This fixes https://github.com/pedestal/pedestal/issues/459

ohpauleez:
Thanks!

grzm:
Thanks for the review!

Issue #456, opened 2016-08-26
Warnings for shadowed functions in Clojure 1.9


grzm:
# Description
When using `org.clojure/clojure "1.9.0-alpha11"`, I see the following warnings due to functions in Pedestal dependencies shadowing clojure.core functions.

```
WARNING: seqable? already refers to: #'clojure.core/seqable? in namespace: clojure.core.incubator, being replaced by: #'clojure.core.incubator/seqable?
WARNING: boolean? already refers to: #'clojure.core/boolean? in namespace: clojure.tools.analyzer.utils, being replaced by: #'clojure.tools.analyzer.utils/boolean?
WARNING: boolean? already refers to: #'clojure.core/boolean? in namespace: clojure.tools.analyzer, being replaced by: #'clojure.tools.analyzer.utils/boolean?
WARNING: bounded-count already refers to: #'clojure.core/bounded-count in namespace: clojure.core.async, being replaced by: #'clojure.core.async/bounded-count
```

## Expected Behavior
I expect to see no warnings. 

## Actual Behavior
I see the following warnings:

```
WARNING: seqable? already refers to: #'clojure.core/seqable? in namespace: clojure.core.incubator, being replaced by: #'clojure.core.incubator/seqable?
WARNING: boolean? already refers to: #'clojure.core/boolean? in namespace: clojure.tools.analyzer.utils, being replaced by: #'clojure.tools.analyzer.utils/boolean?
WARNING: boolean? already refers to: #'clojure.core/boolean? in namespace: clojure.tools.analyzer, being replaced by: #'clojure.tools.analyzer.utils/boolean?
WARNING: bounded-count already refers to: #'clojure.core/bounded-count in namespace: clojure.core.async, being replaced by: #'clojure.core.async/bounded-count
```

## Steps to reproduce
Build a Pedestal app using `[io.pedestal/pedestal.service "0.5.0"]`.

# Environment
```
λ javac -version
javac 1.8.0_45
```

## Operating System (including version).
Mac OS X 10.10.5

## Your current Leiningen version (`lein --version`)
```
λ lein --version
Leiningen 2.6.1 on Java 1.8.0_45 Java HotSpot(TM) 64-Bit Server VM
```
## Pedestal version

`io.pedestal/pedestal.service "0.5.0"`

grzm:
You're likely already aware of these. Including it just to keep track of them.

Looks like some (but not all) of these dependencies already have released versions including the fixes:

## clojure.core.incubator
```
[...]

PR #468, opened 2016-09-20
Bump core.async dependency to 0.2.391

grzm:
This fixes the remaining shadowed function warning mentioned in https://github.com/pedestal/pedestal/issues/456


ohpauleez:
Thanks!

PR #469, opened 2016-09-28
Active streams count should balance.

mtnygard:
Increments were in the loop, being run too many times. Now one increment matches with one decrement.


ohpauleez:
Thanks!

PR #476, opened 2016-10-08
Improve jetty-web-sockets doc

sergioaugrod:
Hello, 

I tried to make the following improvements:

* Add an example of sending messages to all connections
* Send a message only after the connection

Thanks!


ohpauleez:
Hi @sergioaugrod - thanks for helping improve Pedestal!

I'd like to keep the samples as simple/obvious/basic as possible, allowing users to easily use them as a starting point for their own systems.
I like the `send-message-to-all` example function and how you illustrated correctly cleaning up the atom after the client disconnects.  I've incorporated those changes.

Thanks!

PR #478, opened 2016-10-20
Content negotiation should recognize */*

ray1729:
The `negotiate-content` interceptor was not matching `*/*` in the Accept
header. This change updates the `weighted-accept-qs` function to
recognize the wildcard type `*`.

### DIFF
--- a/service/src/io/pedestal/http/content_negotiation.clj
+++ b/service/src/io/pedestal/http/content_negotiation.clj
@@ -40,7 +40,8 @@
 (defn weighted-accept-qs [supported-types accept-elem]
   (when-let [weighted-qs (seq
                              (for [target supported-types
-                                   :let [weighted-q (if (and (= (:type accept-elem) (:type target))
+                                   :let [weighted-q (if (and (or (= (:type accept-elem) (:type target))
+                                                                 (= (:type accept-elem) "*"))
                                                              (or (= (:subtype accept-elem)
                                                                     (:subtype target))
                                                                  (= (:subtype accept-elem) "*")))
--- a/service/test/io/pedestal/http/content_negotiation_test.clj
+++ b/service/test/io/pedestal/http/content_negotiation_test.clj
@@ -38,7 +38,7 @@
 (deftest parse-and-weight-tests
   (testing "single request"
     (are [request-str supported-strs expected]
-         (= (cn/weighted-accept-qs (mapv cn/parse-accept-element supported-strs)
+        (= (cn/weighted-accept-qs (mapv cn/parse-accept-element supported-strs)
                                    (cn/parse-accept-element request-str))
             expected)
 
@@ -67,7 +67,9 @@
             expected)
 
          ;; A multi request; single supported
-         ["foo/bar"] [nil
+         ["foo/bar"] [[55.2
+                       {:params {:q 1.0}, :field "foo/bar", :type "foo", :subtype "bar"}
+                       {:params {:q 0.2}, :field "*/*", :type "*", :subtype "*"}]
                       [105.2
                        {:field "foo/bar" :type "foo" :subtype "bar" :params {:q 1.0}}
                        {:field "foo/*" :type "foo" :subtype "*" :params {:q 0.2}}]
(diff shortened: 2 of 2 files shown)


ohpauleez:
Hi @ray1729!  Thanks for the contribution.  I should have some time to review the changes tomorrow.  In the meantime, have you signed a [Cognitect contributors agreement](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md#contributors-agreement)?

Additionally, I'd like to spend some time understand what problem you're encountering and how this fix address it.
What use case do you have for sending the */* header?  Why would you elect to send the header if you didn't care about the formats coming back?  Since the Pedestal content-negotiation interceptor just provides the decoded/top-ranking format (and doesn't perform any actual decoding), what is the use case for even scoring "*/*" vs just returning nil?

Thanks!

mtnygard:
I think markdown ate the asterisks around your slashes, @ohpauleez

ohpauleez:
A colleague came up with one use case, where you wanted to be very strict about Accept headers and return a 406 if the header was completely missing (vs wildcard type-subtype).  If you other use cases, I'd like to head them.

ray1729:
Hi @ohpauleez , yes I signed the contributor agreement earlier today.

Here's my rationale for this change. I was experimenting with something along the lines of [pedestal-api.content-negotiation](https://github.com/oliyh/pedestal-api/blob/master/src/pedestal_api/content_negotiation.clj) but with (what I thought was) a more idiomatic approach that used  an interceptor to serialize the response body according to the `:accept` key added by Pedestal's `negotiate-content` interceptor. The idea was to use it something like this:
```
(def default-productions
  ["application/json"
   "application/edn"
   "application/transit+json"
   "application/transit+msgpack"
   "application/transit"])

(defn terminally-unacceptable
  [ctx]
  (-> ctx
      (assoc :response (-> (ring-resp/response "Not Acceptable")
                           (ring-resp/status 406)))
      chain/terminate))

(defn api-interceptor
  [f {:keys [produces]
      :or {produces default-productions}}]
  [vary-accept
   (negotiate-content produces {:no-match-fn terminally-unacceptable})
   serialize-response
   f])
```

I was caught out when testing this with an `Accept: */*` header, as the `negotiate-content` interceptor had not populated `:accept` so `serialize-response` did not work as intended.

This use-case aside, I would argue that the current implementation where `*/*` matches nothing is counter-intuitive, as I'd expect it to match anything.

ohpauleez:
Thanks for the feedback @ray1729 

In the system/use-case you posted, if a client didn't set an Accept header, you'd reject the request with a 406, correct?  Is that the intended behavior you'd take into production?

ray1729:
Thank you for prompting me to think about the case where the client does not send an Accept header. In my application, I'd like to treat it as if the client sent `Accept: */*`. I'm not sure if you'd want to encode that behaviour in `negotiate-content` - perhaps as an option? 

From my reading of the code, if there's no Accept header the `negotiate-content` interceptor would leave the context unchanged, and in that case I could interpret nil `:accept` as "not specified" and handle this case with a default choice in my `serialize-response` interceptor.

mtnygard:
I think the pertinent part of RFC 2616 is this:

> If no Accept header field is present, then it is assumed that the client accepts all media types. If an Accept header field is present, and if the server cannot send a response which is acceptable according to the combined Accept field value, then the server SHOULD send a 406 (not acceptable) response.

From https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1.

ohpauleez:
Given the RFC, I'm going to close this PR.  An accept header of `*/*` can be treated the same as no accept header.  Looking up the accept score and seeing `nil` should mean "Accepts all types".  If someone disagrees, please comment and we can open up a new issue.

Thanks everyone!

ray1729:
@ohpauleez Perhaps I didn't explain the problem fully. As I understand it, Pedestal's *current* behavior is contrary to the RFC. Starting a REPL in a checkout of the current master branch:
```
user=> (require 'io.pedestal.http.content-negotiation-test)
nil
user=> (in-ns 'io.pedestal.http.content-negotiation-test)
#object[clojure.lang.Namespace 0xf48915d "io.pedestal.http.content-negotiation-test"]
io.pedestal.http.content-negotiation-test=> (-> (context {:uri "/blah" :headers {"accept" "*/*"}}) (app) ((:enter (cn/negotiate-content ["application/json"]))))
{:request {:headers {"accept" "*/*"}, :request-method :get, :uri "/blah"}, :response {:status 406, :body "Not Acceptable"}}
```
The pull request I submitted fixes this.

ohpauleez:
Sorry about that @ray1729 -- I rarely use the default interceptor and usually just roll the specific behavior I'm looking for with the available scoring functions.  Have you signed the Cognitect contributors agreement I linked above?

ray1729:
@oupauleez yes, signed electronic version on 10/20/2016, are you able to check it was received by Cognitect?

PR #480, opened 2016-11-03
Bugfix definterceptor macro.

ray1729:
I ran into a problem with the `definterceptor` macro:

```
$ cd pedestal/interceptor
$ lein repl
user=> (require '[io.pedestal.interceptor.helpers :refer [definterceptor]])
user=> (definterceptor foo {:name ::foo :enter identity})

AssertionError Assert failed: (valid-interceptor? %)  io.pedestal.interceptor/interceptor (interceptor.clj:78)
```

But this works:

```
user=> (def m {:name ::foo :enter identity})
#'user/m
user=> (definterceptor foo m)
#'user/foo
```

It looks like a small bug in the unquoting in the `definterceptor` macro.  What do you think to this proposed fix?

### DIFF
--- a/interceptor/src/io/pedestal/interceptor/helpers.clj
+++ b/interceptor/src/io/pedestal/interceptor/helpers.clj
@@ -34,7 +34,7 @@
                (first body))
         doc (when (string? (first body))
               (first body))]
-    `(def ~(with-meta name {:doc doc})  ~(interceptor init))))
+    `(def ~(with-meta name {:doc doc})  (interceptor ~init))))
 
 (defn- infer-basic-interceptor-function
   "Given list `args`, infer a form that will evaluate to a function


ohpauleez:
Hi @ray1729!  Thanks for submitting a possible fix.

All of those macros should be avoided, are considered deprecated, and are only still there to maintain backwards compatibility for older Pedestal code bases.

You should always prefer to use the [Interceptor API](https://github.com/pedestal/pedestal/blob/master/interceptor/src/io/pedestal/interceptor.clj#L78) directly.  There's no longer a need for any of the macro-based builders.
For more information, see the [releases](https://github.com/pedestal/pedestal/releases) or the [Interceptor API PR](https://github.com/pedestal/pedestal/pull/301).

ray1729:
Hi @ohpauleez that's fine, I already switched to using `io.pedestal.interceptor/interceptor` directly. But just to be clear, the bug is in `io.pedestal.interceptor.helpers/definterceptor`, not the deprecated `definterceptorfn`. Are you saying that these new helpers are now also deprecated?

ohpauleez:
Yep, you shouldn't be using any of the macros unless you're porting from an existing (older) Pedestal code base.  They also prevent AOT from working correctly.

If the bug is occurring from you porting over an older Pedestal service, let me know and we'll reopen.

ray1729:
No, this is a new project, so I'll steer clear of these macros. Thank you for the feedback.

Issue #488, opened 2016-12-19
Malformed request can trigger an internal server error


ray1729:
# Description

Incorrectly encoded URL parameters trigger an internal server error.

## Expected Behavior

Server should respond with 400 - bad request.

## Actual Behavior

Server responds with 500 - internal server error.

## Steps to reproduce

Run the hello world application from the Pedestal guide: http://pedestal.io/guides/hello-world

Issue the request `curl http://localhost:8890/greet?q=%`

# Environment

## Operating System (including version).

Ubuntu trusty 

## Your current Leiningen version (`lein --version`)

Leiningen 2.7.1 on Java 1.8.0_77 Java HotSpot(TM) 64-Bit Server VM

## Pedestal version

0.5.1

ohpauleez:
On master.  Thanks!

PR #489, opened 2016-12-19
Catch URLDecoder error, return bad request response.

ray1729:
Putting this forward as a possible fix for #488 

I'm not 100% happy with this approach, so would be interested to know if you have a better suggestion.

### DIFF
--- a/route/src/io/pedestal/http/route.clj
+++ b/route/src/io/pedestal/http/route.clj
@@ -64,12 +64,7 @@
   "Decodes one key or value of URL-encoded UTF-8 characters in a URL
   query string."
   [string]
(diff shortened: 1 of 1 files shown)


ohpauleez:
Thanks for jumping in and putting something together!

I would drop the exception handling around URLDecode (since it's not adding much), and then just rewrite the exception handling in `query-params` to catch and package up a 400 response.

ray1729:
Thanks  for the feedback. Updated to simplify exception handling.

ohpauleez:
Thanks for the proposed changes.  I rolled them into master with a test to confirm they work as expected.

PR #496, opened 2017-02-26
Multimethods are now first class handlers.

marctrem:
I believe this is convenient!


ohpauleez:
Thanks for participating @marctrem!

I would rather leave MultiFn out of the core/forced protocol extension block.  Some applications use multimethods for Context control and others use it for response generation.  Leaving it out means that each application can `extend-protocol` in their application code and choose the implementation that makes the most sense.

PR #509, opened 2017-04-21
Fixed `run-dev` in chain-providers sample.

ddeaguiar:
- The sample chain provider does not use routes.
- Implemented a working test.


ddeaguiar:
Changes on master.

PR #514, opened 2017-05-22
service-template - Add heroku support

rhacker:


mtnygard:
Reviewed: 👍

ohpauleez:
I started digging into this more.  In addition to changing the the way environment variables are handled, I'd also prefer to launch instances using an uberjar instead of the Leiningen trampoline.  Originally I was looking to see if we could just use the docker setup when I saw that Heroku's docs actually suggest using uberjars over trampolining leiningen -- https://devcenter.heroku.com/articles/clojure-support#uberjar

ddeaguiar:
Closing this due to no activity. It can be revisited if/when Heroku comes up again.

PR #515, opened 2017-05-30
Pass event channel into context

fl00r:
We would like to close it in `on-client-disconnect`. It is the only piece of communication with `stream-ready-fn` thread.


ohpauleez:
For others who see this via search:

The event-channel and the response-channel are both closed before `on-client-disconnect` is called.  `on-client-disconnect` is typically used for cleaning up app-specific bookkeeping

fl00r:
@ohpauleez thank you. I hurried. I found it out after I've created the issue.

Issue #499, opened 2017-03-02
Pedestal breaks my ClojureScript / JavaScript application


ohpauleez:
# Description

When using Pedestal, my application works fine in Firefox (albeit some warnings), but fails to even load in Chrome.

## Expected Behavior

When serving up my JavaScript application from the resource/public path in Pedestal, I expect my application to load in the browser

## Actual Behavior

A bunch of CSP errors/warnings within the JavaScript console and the application never loads

## Steps to reproduce

Any JavaScript or ClojureScript application that uses dynamic features (forced writing, eval, etc.)

## Pedestal version

0.5.2

ohpauleez:
In the latest version of Pedestal, Content Security Policy headers are set by default.  This was documented in the [release notes for 0.5.2](https://github.com/pedestal/pedestal/releases/tag/0.5.2) and issue #466 .

There are [many options for Content Security Policy](https://content-security-policy.com/) and Pedestal has attempted to choose sane defaults that offer good protection.

Nonetheless, sometimes you have a browser application (or are using a JavaScript library) that needs dynamic behavior.

In the service map, you can set Content Security Policy settings that work best for you.  Given Pedestal's other secure headers and CSRF protection, you may choose something as basic as:

```clojure
(def service
  {...
   ::http/secure-headers {:content-security-policy-settings {:object-src "none"}
  ... }
```

PR #521, opened 2017-06-17
Relax Content Security Policy in dev map in Leiningen template

eneroth:
Add the solution from issue #499 to the Leiningen template.


eneroth:
Used the wrong ns name from the issue vs what's in the template.

PR #522, opened 2017-06-17
Relax Content Security Policy for dev mode in Leiningen template

eneroth:
Add the solution from issue #499 to the Leiningen template.

### DIFF
--- a/service-template/src/leiningen/new/pedestal_service/server.clj
+++ b/service-template/src/leiningen/new/pedestal_service/server.clj
@@ -20,7 +20,9 @@
               ;;  we can use this to set the routes to be reloadable
               ::server/routes #(route/expand-routes (deref #'service/routes))
               ;; all origins are allowed in dev mode
-              ::server/allowed-origins {:creds true :allowed-origins (constantly true)}})
+              ::server/allowed-origins {:creds true :allowed-origins (constantly true)}
+              ;; Relax Content Security Policy (CSP) in dev mode
+              ::server/secure-headers {:content-security-policy-settings {:object-src "none"}}})
       ;; Wire up interceptor chains
       server/default-interceptors
       server/dev-interceptors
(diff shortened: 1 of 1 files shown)


ohpauleez:
Thanks for using Pedestal and offering up a contribution!

I think this change may end up causing more issues for users down the line (certain things would "work" in dev mode, but fail to work correctly in production).  Rather than relax the CSP policy in dev mode, what if we added the option to the service map in `service.clj`, commented out, with a comment explaining why you might want to use it with links from #499 

I think we have an opportunity to educate the users about CSP instead of just masking over the details (and I'd like the default policy to be in effect, as-is, in dev and prod mode).

eneroth:
I ran into the problem of Chrome blocking app.js, as generated by the ClojureScript compiler. The reason is because of these lines:
```
var CLOSURE_UNCOMPILED_DEFINES = null;
if(typeof goog == "undefined") document.write('<script src="js/out/goog/base.js"></script>');
document.write('<script src="js/out/cljs_deps.js"></script>');
document.write('<script>if (typeof goog == "undefined") console.warn("ClojureScript could not load :main, did you forget to specify :asset-path?");</script>');

document.write("<script>if (typeof goog != \"undefined\") { goog.require(\"figwheel.connect.build_client_dev\"); }</script>");
document.write('<script>goog.require("client.dev");</script>');
```

So ClojureScript intentionally generates unsafe js when in development mode. However, this does not happen in production mode. For this reason, it's a necessity to relax CSP when in dev mode on Pedestal, and I don't think it's a good idea to flip this switch back and forth manually. You probably want to leave security firmly ON in prod mode, while still avoiding having Chrome block your ClojureScript code while in dev mode.

eneroth:
One possible way to make this distinction more obvious is to explicitly set the CSP to `strict-dynamic`, along with a comment as to why, in the production map over in `service.clj`, even though this is the default setting for Pedestal. Just to make the difference explicit.

ohpauleez:
This is addressed on master now.  CSP is relaxed in dev mode, and the service map has instructions (and an example) on how to toggle CSP accordingly.  Thanks for the suggestions and conversation!

PR #530, opened 2017-07-17
log-request interceptor made optional.

lxsameer:
Problem:
Not every one need the `log-request` interceptor. Since pedestal
is a library it should give the control to user to customize every
aspect of the system ( at least as much as possible ).

Solution:
`log-request` made optional. By default it would be the first interceptor
in the vector returned `default-interceptors`. But if the `::http/no-logging`
key had a truthy value in the service map then log request interceptor
won't be conjed into the default interceptors vector.

Personally I have my own logging interceptor which fits my needs but `log-request` interceptor is something that don't help me in my case but i didn't have the chance to disable it. I either had to create the default interceptor stack manually or do some hack to remove it from system map

### DIFF
--- a/service/src/io/pedestal/http.clj
+++ b/service/src/io/pedestal/http.clj
@@ -222,6 +222,7 @@
          enable-session ::enable-session
          enable-csrf ::enable-csrf
          secure-headers ::secure-headers
+         no-logging ::no-logging
          :or {file-path nil
               router :map-tree
               resource-path nil
@@ -230,6 +231,7 @@
               ext-mime-types {}
               enable-session nil
               enable-csrf nil
+              no-logging false
               secure-headers {}}} service-map
         processed-routes (cond
                            (satisfies? route/ExpandableRoutes routes) (route/expand-routes routes)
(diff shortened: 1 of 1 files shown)


ohpauleez:
Hi @lxsameer - thanks for the suggestion and pulling the PR together.  I made a few changes to your idea and pushed the changes to master.

For production Pedestal services, I usually recommend that users define their own interceptor chain instead of using the default chain -- it certainly seems like you're at that level of experience/use/expertise.
That said, I think there are also users who stick to the default interceptors regardless.  Exposing some additional options/flexibility in configuring the default interceptors would result in a better user experience.

Since all of the options are in the affirmative, I switched your `:no-logger` to `:request-logger` - allowing users to set it to `nil` to disable it, or pass in any custom request logger they like (just needs to satisfy the IntoInterceptor protocol).  Related, I also setup `not-found-interceptor` the same way - nil disables the interceptor, otherwise you can use anything that can be turned into an interceptor (previously you had to pass in a fully created interceptor).

Hope this changes meet your needs!

PR #531, opened 2017-07-21
replace some uses of deprecated defroutes, fix sample 'chain-providers'

tbcs:


ohpauleez:
Wow, awesome!  Thanks so much.  Let me dig into this a bit so we can get it merged.

bonkydog:
Hello, @tcbs!  

Thanks for putting time and attention into making Pedestal better!

I ran into a problem with this patch: replacing (the deprecated) `defroutes` macro with `(def routes `[... TERSE ROUTE HERE ...])` or `(def routes `#{ ... TABLE ROUTE HERE...})` breaks the example's dev server.

Before, the `defroutes` macro would expand these convenient route formats into the verbose "real" route structure used by the router. Now, omitting the call to `expand-routes` allows the routes to remain a "terse" route (a vector of vectors) or a "table" route (a set).

That's OK for the example's "production" server, where the route is defined and then used to set up the service once, but we run into problems in the "dev" server. We want to enable "hot-reloading" the routes, so we wrap the routes in a function that derefs the routes var for each request.

But oh no! Here's how [default-interceptors](https://github.com/pedestal/pedestal/blob/b9610e6d27bb88db985082cb9f21176a4001b33a/service/src/io/pedestal/http.clj#L239) treats data structures that have an implementation for ExpandableRoutes protocol (that is, vectors of vectors for terse routes, sets for table routes) versus how it treats routes that have been wrapped in a function:

```clojure 
(cond
    (satisfies? route/ExpandableRoutes routes) (route/expand-routes routes)
    (fn? routes) routes
    (nil? routes) nil
    (and (seq? routes) (every? map? routes)) routes
    :else (throw (ex-info "Routes specified in the service map don't fulfill the contract.
                         They must be a seq of full-route maps or satisfy the ExpandableRoutes protocol"
                         {:routes routes})))
```

Function-wrapped routes don't get expanded. (They'd need to be expanded every request.)

So replacing

`(defroutes routes [... TERSE ROUTE ...])`

with

`(def routes `[... TERSE ROUTE ...])`

breaks the dev servers included in the examples.

So what do we do here? 

We could call `expand-routes` when the route is defined:

```clojure
(def routes
  (route/expand-routes
   `[[["/"
       ["/hello" {:get hello-world}]]]]))
```

OR

We could call `expand-routes` in the wrapping function:

```clojure
  {
   ;; ...
   ::server/routes #(route/expand-routes (deref #'service/routes))
   ;; ...
   }
```

Which do you think is less confusing for someone new to Pedestal (and maybe Clojure)?

ddeaguiar:
I think we should follow what's in the [service template](https://github.com/pedestal/pedestal/blob/50fe5ea89108998ac5a79a02a44432fd111ea6f8/service-template/src/leiningen/new/pedestal_service/server.clj#L21) where `expand-routes` is called explicitly in the `run-dev` implementation.

bonkydog:
OK, will fix and merge soon.

bonkydog:
OK, fixed the dev servers and merged from master to this branch to pre-resolve merge conflicts.

bonkydog:
Currently testing the examples...

bonkydog:
OK, this is ready. Samples all run.

hlship (approved):
Side discussion on Slack (https://nubank.slack.com/archives/C018QBP6VHD/p1663187392040359) addresses what's going on with the extra levels of indirection (derefs and back ticks) and how it should be documented.

PR #540, opened 2017-10-20
Content negotiation: fix the default no-match-fn

zaninime:
The default function does not return a valid ring response (checked here https://github.com/pedestal/pedestal/blob/41337e41ae773f3c17baeb1c244be13ddf6c28db/service/src/io/pedestal/http/impl/servlet_interceptor.clj#L233), so the rest of the interceptors are called even when the content negotiation fails.

Adding `:headers {}` to the response map solves it.


ohpauleez:
This fix is now on master, thanks!

PR #544, opened 2017-12-18
Fix pre-condition in chain/enqueue

ordnungswidrig:
Small fix for the precondition declaration on chain/enqueue

### DIFF
--- a/service/test/io/pedestal/interceptor_test.clj
+++ b/service/test/io/pedestal/interceptor_test.clj
@@ -196,11 +196,33 @@
                                  (catcher :h)])
                        :leave))))
 
-(deftest t-enque-precondition
+(deftest t-enqueue
   (is (thrown? AssertionError
-               (execute (enqueue {} [nil]))))
+               (enqueue {} [nil]))
+      "nil is not an interceptor")
   (is (thrown? AssertionError
-               (execute (enqueue {} [(fn [_])]))))) 
+               (enqueue {} [(fn [_])]))
+      "function is not an interceptor")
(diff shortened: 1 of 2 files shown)


ohpauleez:
I'm not sure why the service-template tests are failing on this change, but all tests should be fixed before this gets merged.  Additionally, we should add tests that exercise the precondition.

ddeaguiar:
@ohpauleez the template tests are failing due to an issue with lein version 2.8.1. Refer to #558 which has been merged to master.

ordnungswidrig:
@ohpauleez I'll happily add some tests but the interceptor project is lacking tests alltogether. Is there a separate place for the tests?

ohpauleez:
@ordnungswidrig All tests for the entire project live in `pedestal.service`.  Probably makes the most sense to put them in the [base interceptor test file](https://github.com/pedestal/pedestal/blob/master/service/test/io/pedestal/interceptor_test.clj)

Thanks all!

ordnungswidrig:
@ohpauleez test added.

service/test/io/pedestal/interceptor_test.clj
```diff
@@ -196,6 +196,12 @@
                                  (catcher :h)])
                        :leave))))
 
+(deftest t-enque-precondition
```
ddeaguiar: @ordnungswidrig can you add a test with a valid interceptor and drop the call to `execute`? It's not needed to test the precondition.

ddeaguiar:
@ordnungswidrig looks good. Merging, thanks :)

PR #546, opened 2018-02-14
Expand the async example and explanation

hlship:
Was reading through the docs and the async example bothered me, since it appeared to show a long-running process inside a go block. I touched that up, simplified it with an explanation of what go and <! does, and otherwise expanded the description.


hlship:
Killing this, and submitting a similar PR to pedestal-docs.

ohpauleez:
It's like you read my mind.  Thanks so much @hlship!

PR #548, opened 2018-03-16
SSE: Properly decrement io.pedestal.http.sse/active-streams metric

milt:
Moved the call to decrement the active streams counter metric outside of the dispatch loop, after the `on-client-disconnect` callback. Previously, the counter would not decrement if the stream was closed on a heartbeat or send failure.


ohpauleez:
Thanks for taking the time to pull together a PR.  I'll have to take some time to look into this and see if we can also get a test in place.  A small change before this gets accepted: The last expression should be the call to `on-disconnect`, with the decrement happening just before it.

milt:
👍 Sounds good.

ohpauleez:
This fix is now on master, thanks!

PR #551, opened 2018-04-06
Add a message diagnostic context to io.pedestal.log

hlship:
This small change adds a message diagnostic context as a dynamic Var.  Key values in the MDC are added to every logged event.

Although some underlying logging implementation have an MDC, it is thread-bound and unaware of Clojure per-thread bindings, which is especially significant w.r.t. core.async.

A typical use of the MDC is to incorporate a request correlation id into all logged events.

### DIFF
--- a/log/src/io/pedestal/log.clj
+++ b/log/src/io/pedestal/log.clj
@@ -109,6 +109,11 @@
     ([t body] nil)
     ([t body throwable] nil)))
 
+(def ^:dynamic ^{:added "0.5.4"} *mdc*
+  "The MDC (message diagnostic context) is a map of additional keys and values to be added
+  to every event. Per-event values may override the MDC."
+  nil)
+
 ;; Override the logger
 ;; ---------------------
 ;; Pedestal's logging is backed by a protocol, which you are free to extend
@@ -119,7 +124,7 @@
 ;; JVM Property 'io.pedestal.log.overrideLogger' or ENVAR 'PEDESTAL_LOGGER'
 ;; to a symbol that resolves to a single-arity function
 ;; (passed a string logger tag, the NS string of the log call).
-;; This function should return something that satisifes the LoggerSource protocol.
+;; This function should return something that satisfies the LoggerSource protocol.
 ;; The function will be called multiple times (as the logging macros are expanded).
 
 (defn- log-expr [form level keyvals]
@@ -137,9 +142,10 @@
                          `(LoggerFactory/getLogger ~(name (ns-name *ns*))))]
        (when (io.pedestal.log/-level-enabled? ~logger' ~level)
          (let [~string' (binding [*print-length* 80]
-                          (pr-str (assoc (dissoc ~keyvals-map
-                                                 :exception :io.pedestal.log/logger)
-                                         :line ~(:line (meta form)))))]
+                          (-> (merge *mdc* ~keyvals-map)
+                              (dissoc :exception :io.pedestal.log/logger)
+                              (assoc :line ~(:line (meta form)))
+                              pr-str))]
            ~(if exception'
               `(~log-method' ~logger'
                              ~(with-meta string'


hlship:
I'm pretty sure the build was broken before this PR; I had build errors locally.

ddeaguiar:
@hlship the build issue seems to be related to lein 2.8.1. I'm looking into it.

hlship:
I got the error building locally, and I'm still on lein 2.7.1.

ddeaguiar:
@hlship that is strange. Here's what I get when I run the tests as per the ci build from the service dir:

```
$ lein version ; lein test :travis
Leiningen 2.7.1 on Java 1.8.0_112 Java HotSpot(TM) 64-Bit Server VM

lein test pedestal.new-server-integration-test

Ran 0 tests containing 0 assertions.
0 failures, 0 errors.
```

```
$ lein2 version; lein2 test :travis
zsh: correct 'lein2' to 'lein' [nyae]? n
Leiningen 2.8.1 on Java 1.8.0_112 Java HotSpot(TM) 64-Bit Server VM
clojure.lang.Compiler$CompilerException: java.lang.NullPointerException, compiling:(new_server_integration_test.clj:22:3)
 at clojure.lang.Compiler$InstanceMethodExpr.eval (Compiler.java:1532)
    clojure.lang.Compiler$InvokeExpr.eval (Compiler.java:3651)
    clojure.lang.Compiler$InstanceMethodExpr.eval (Compiler.java:1517)
...
```

Notice that the tests are meant to be excluded in the ci build. The failure is due to `*file*` being nil when the tests are run with lein 2.8.1.

hlship:
This has been stewing for a bit ... anything I can do on my end?

ddeaguiar:
@hlship, sorry for the delay. We've pushed the fix for the build failures to master. Can you update your PR with it? I'll take a closer look at it over the weekend. Thanks!

log/src/io/pedestal/log.clj
```diff
@@ -109,6 +109,11 @@
     ([t body] nil)
     ([t body throwable] nil)))
 
+(def ^:dynamic ^{:added "0.5.4"} *mdc*
```
ddeaguiar: `MDC` is already well known as _Mapped Diagnostic Context_ in logging implementations. Using a similar abbreviation here is confusing, particularly when since the semantics differ.

ohpauleez:
I've found it pretty common to make my own app-specific logging functions, which get added into the context by an interceptor.  This most often happens because of log aggregation services and formatting rules, but I've also done it to always log trace/request IDs in log messages.  I would rather not introduce a dynamic var into core of the logging function, and instead let people control the logger/logging functions however they see fit.

Distributed trace should land in `log` soon, which might reduce the need of this specific change to some degree.

All of that said, I'm very interested to hear input from all users!  @hlship How are you currently working around this and what is particularly painful that we might do better?

log/src/io/pedestal/log.clj
hlship: Calling it the message diagnostic context may have just been a slip of the tongue.

It's possible to use the MDC from Logback BUT that has two problems:
- The added data is outside the EDN map emitted by pedestal.log
- Coordinating the per-thread MDC data in Logback against multiple threads (especially with core.async) is problematic

The solutions we've used for the MDC data usually ended up involving some leaked data; the MDC would be updated in a thread and that data might persist after the logging on that thread completed. So you had to be very careful to initialize all the potential keys of the MDC early in each request pipeline, which is quite haphazard.

We don't currently use pedestal.log in our application production code, we use our own wrappers around Logback. This is for some open source code I'm working on, but if it goes through I can start looking into moving our production code to pedestal.log.

ohpauleez:
@hlship I landed some MDC stuff on master, including a dynamic var you can use however you see fit.  [Take a look](https://github.com/pedestal/pedestal/blob/master/log/src/io/pedestal/log.clj#L332) and if that meets your needs, you can close this PR.  Thanks!

hlship:
Although having some way to put data into the SLF4J MDC is nice, it still means that we have two separate sets of data: the Pedestal logging map, and the SLF4J MDC (which, effectively, contains a map in one of its keys), and the SLF4J MDC, being defined outside of Clojure, can only honor the MDC changes within a single Clojure body.

Supporting a io.pedestal.log specific MDC integrates the data together in a single map (which is by far my preference); it will be inherited by any `go` or `thread` blocks spun up from inside an interceptor, and we can use the existing interceptor mechanism to propagate the binding into later interceptors in the pipeline.

So, no, the solution that dropped in master is not useful.

Also, I don't understand why in that `with-context` macro, you don't use `-get-mdc` and store that in a `let` scoped symbol; that seems like the best way to ensure that the io.pedestal MDC key is restored to it's previous value (honoring the previous formatter, if
different). Feels like more straight forward code.

ohpauleez:
Let me walk through the changes that exist on master.

They achieve the following:
 * Introduce a new protocol for MDC -- and you're free to interpret that however you want to, including making your own app-specific MDC, enabling the use case you originally wanted
 * Bridge the MDC protocol to SLF4J's MDC interface
 * Introduce a dynamic var (`*mdc-context*`) for performing Clojure-specific or Clojure-advantageous MDC-like things.  No central logging code depends on this, you can take full control if you like.
 * Introduce two macros that achieve the same thing with different amounts of control and overhead -- Using dynamic scope to manage a Clojure-specifc "MDC" which is stored in a single key within an MDC (anything that works with the protocol).

Aside from the MDC-related changes, fn-based logging and distributed tracing are also on master now.
I think they complement the standard logging and metrics quite well.

It is intended that all of these tools/utilities complement each other, but remain separate.  It is also my hope they they provide the necessary building blocks for people to build necessary app-specific functionality.

ohpauleez:
I'm going to close this PR for now, but I'm happy to continue the discussion elsewhere.  The 0.5.4 release provides some base-level utilities in this area, and `io.pedestal.log` provides the foundation needed for applications and services to tailor custom/specific functionality.  I think using the dynamic context var within logging message maps (if desired) goes a long way.

Issue #520, opened 2017-06-14
Move Servlet interceptor chain provider specifics to servlet-specific namespaces


ddeaguiar:
# Description
The implementation of HTTP-based, non-servlet interceptor providers is hampered by the inclusion of Servlet-specific implementation details in the following namespaces: [io.pedestal.http.request](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/request.clj#L13), [io.pedestal.http.ring-middlewares](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/ring_middlewares.clj#L34), [io.pedestal.http.request.map](https://github.com/pedestal/pedestal/blob/f89d0abe244d0e8385d66a697724ef56180a00e5/service/src/io/pedestal/http/request/map.clj#L3).

## Expected Behavior
The implementation of HTTP-based, non-servlet interceptor providers should be able to benefit from the request/response abstractions, processing utilities and generic ring middleware implementations that Pedestal provides.

## Proposed Solution
* Remove the type hint from the [io.pedestal.http.request.map/fast-resource](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/ring_middlewares.clj#L168) implementation.
* Move the servlet-specific bits in the `io.pedestal.http.request` and `io.pedestal.http.request.map` namespaces to the following new namespaces`io.pedestal.http.servlet.request` and `io.pedestal.http.servlet.request.map`, respectively.

## Pedestal version
0.5.2

ddeaguiar:
@ohpauleez, I'd be happy to implement the PR for the approach we decide on taking.

ohpauleez:
Thanks @ddeaguiar for opening up the issue.  For the time being, you could just include the Servlet API in your dependencies and these issues would go away.

- - - - 

### Proposed solutions (revised)

 * Move the Servlet details out of `request.clj` and into `request/servlet_support.clj` -- where we'll extend the protocol to the ServletRequest type and put the servlet request parsing utils.  We'll have to update the `:require` references to `request.clj` in a few places.
 * Leave `request/map.clj` as is, with the Servlet details in play.  Move the `base-request-map` function up into `request.clj`.  Often times, custom chain-providers end up specifying their own "request" implementation anyway (so I don't think there's really a reuse issue here).
[...]

PR #553, opened 2018-04-14
Move servlet-specifics out of more general http namespaces

ddeaguiar:
This PR addresses #520, following the implementation approach provided by @ohpauleez. Refer to the issue for details.

### DIFF
--- a/service/src/io/pedestal/http/request/servlet_support.clj
+++ b/service/src/io/pedestal/http/request/servlet_support.clj
@@ -10,8 +10,8 @@
 ; You must not remove this notice, or any other, from this software.
 
 (ns io.pedestal.http.request.servlet-support
-  (:require [io.pedestal.http.request :as http.request])
-  (:import (javax.servlet.http HttpServletRequest)))
+  (:require [io.pedestal.http.request :as request])
+  (:import (javax.servlet.http HttpServletRequest HttpServletResponse)))
 
 (defn servlet-request-headers [^HttpServletRequest servlet-req]
   (loop [out   (transient {})
@@ -30,7 +30,7 @@
       "/"
       path-info)))
 
-(extend-protocol http.request/ContainerRequest
+(extend-protocol request/ContainerRequest
   HttpServletRequest
   (server-port [req] (.getServerPort req))
   (server-name [req] (.getServerName req))
@@ -47,3 +47,7 @@
   (path-info [req] (servlet-path-info req))
   (async-supported? [req] (.isAsyncSupported req))
   (async-started? [req] (.isAsyncStarted req)))
+
+(extend-protocol request/ResponseBuffer
+  HttpServletResponse
+  (response-buffer-size [req]))
--- a/service/src/io/pedestal/http/request/servlet_support.clj
+++ b/service/src/io/pedestal/http/request/servlet_support.clj
@@ -50,4 +50,5 @@
 
 (extend-protocol request/ResponseBuffer
   HttpServletResponse
-  (response-buffer-size [req]))
+  (response-buffer-size [resp]
+    (.getBufferSize ^HttpServletResponse resp)))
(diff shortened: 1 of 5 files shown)


service/src/io/pedestal/http/request/servlet_support.clj
```diff
@@ -0,0 +1,53 @@
...
+  (header [req header-string] (.getHeader req header-string))
+  (ssl-client-cert [req] (.getAttribute req "javax.servlet.request.X509Certificate"))
+  (body [req] (.getInputStream req))
+  (path-info [req] (servlet-path-info req))
+  (async-supported? [req] (.isAsyncSupported req))
+  (async-started? [req] (.isAsyncStarted req)))
+
+(extend-protocol request/ResponseBuffer
+  HttpServletResponse
+  (response-buffer-size [req]))
```
ohpauleez: You still need to move `.getBufferSize` in here, correct?

ddeaguiar: Yep, missed that :/. I'll rename the param to `resp` as well.

ohpauleez:
Thanks a ton for pulling all these changes together -- they represent all the changes I had in mind.

Going forward, you can just make branches on the Pedestal repo directly (it makes looking at diffs locally a little easier).

One thing that I'd typically do in a changeset like this is take some good profiling and benchmark numbers before and after.  To do that, I look at the following:
 * Profiling info from common/happy path after JIT warm-up (do the changes prevent JIT optimizations? Where are we spending most of our time?  What calls are dominating?  Is there a clear hot spot that needs to be addressed?)
 * Length of the test suite running
 * Throughput/goodput and latency across standard applications (I have a handful of Pedestal applications I test, but you could always use the samples as a stand in if you don't have applications on hand)
 * Throughput/goodput and latency of the fastest Pedestal setups (again, I have some internal stuff I use).
 * Any other benchmarks or inspections that make sense given the changes

ddeaguiar:
@ohpauleez, Good point. In regards to the fastest Pedestal setups, would the [fast-pedestal](https://github.com/pedestal/pedestal/tree/master/samples/fast-pedestal) sample be sufficient or should I consider other setups as well?

On a separate note, It's not clear to me what's causing the integration test failure. It passes on my machine but I noticed that I'm using a different version of Java (1.8.0_112) and lein (2.7.1). Perhaps it's related to that? I'll try upgrading locally to verify.

Thanks for taking the time to review this!

ddeaguiar:
@ohpauleez, I had tried to create the branch in this repo but I don't have access yet. If you or @mtnygard can grant me the requisite privileges I'll do that in the future.

ohpauleez:
@ddeaguiar `fast-pedestal` sample would be sufficient, in both the `fast-jetty` and `faster-jetty` setup, using both `wrk` and `wrk2` for getting numbers.  If you have a real Pedestal project that uses a highly optimized chain provider (like a service router or something), that's also usually a better source for comparative testing because it actually performs some sort of business logic.

I'll let @mtnygard sort out the perms for you.

Ping me here if you need an extra set of eyes to sort out the integration issue.

ddeaguiar:
@ohpauleez, the ci builds are failing because the server-integration-test [depends on *file*](https://github.com/pedestal/pedestal/blob/master/service-template/test/pedestal/new_server_integration_test.clj#L23) which is nil with lein 2.8.1. Ironically these tests are not intended to be run in the ci build (hence the usage of `lein test :travis`). I'll spend some time looking into this but I'm open to suggestions.

ddeaguiar:
Ping @mtnygard, @ohpauleez. I updated with `master` and CI is now green. I ran some benchmarks with `wrk` and `wrk2` against the `faster-jetty` sample and did not observe a performance impact. I think this PR is ready to merge.

I'm going to look into automating the performance test verification.

ohpauleez:
I'm merging this.  It'll clean up a handful of branches I've got in-flight.

Issue #547, opened 2018-02-25
:path-params are not url decoded


patrickthebold:
In the :request map the :query-params are url decoded, but the :path-params are not. This seems oddly inconsistent.
For example I was testing with random data and expected:
```
PUT /user/u%3A%27%24%5D%3FK%40%3As%7DC1%5B07A%27FQQ%3FBE%2F%29%094G%23%2Ci_%0A-%3D%3F-v%29b%22%24G%3FG%5D%3B%24%3B%26M%60L%3DX%5E-z%3ASENn
```
to be decoded.
If I put the same data in the query string, the string is decoded.

ohpauleez:
I agree that all parts of the url should be url-decoded.  Thanks for reporting.
If someone picks this ticket up, please include a handful of tests.

ohpauleez:
Fixed in #559 -- Thanks!

PR #559, opened 2018-05-08
decode path params interceptor & tests

joakimmohn:
This PR addresses #547. Let me know if I missed something.

### DIFF
--- a/route/src/io/pedestal/http/route.clj
+++ b/route/src/io/pedestal/http/route.clj
@@ -121,11 +121,20 @@
                     key
                     (.append b c)))))))))
 
+(defn parse-param-map [m] (reduce
+                           (fn [acc [key val]] (assoc acc key (decode-query-part val)))
+                           {}
+                           m))
+
 (defn parse-query-params [request]
   (merge-with merge request
               (when-let [string (:query-string request)]
                 (let [params (parse-query-string string)]
                   {:query-params params :params params}))))
+
+(defn parse-path-params [request]
+  (let [m (:path-params request)]
+    (assoc request :path-params (parse-param-map m))))
 
 ;;; Combined matcher & request handler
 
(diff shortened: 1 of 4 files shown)


route/src/io/pedestal/http/route.clj
```diff
@@ -121,12 +121,21 @@
                     key
                     (.append b c)))))))))
 
+(defn parse-param-map [m] (reduce
```
ohpauleez: This should build up a transient and potentially use `reduce-kv` to remove some overhead.

route/src/io/pedestal/http/route.clj
```diff
@@ -121,12 +121,21 @@
...
+                           m))
+
 (defn parse-query-params [request]
   (merge-with merge request
               (when-let [string (:query-string request)]
                 (let [params (parse-query-string string)]
                   {:query-params params :params params}))))
 
+(defn parse-path-params [request]
+  (let [m (:path-params request)]
```
ohpauleez: This needs to be an `if-let` to remove adding a potential `nil` entry.

service/test/io/pedestal/http/route_test.clj
```diff
@@ -955,6 +955,19 @@
        "/?:@-._~!$'()*+,;=/?:@-._~!$'()*+,;=="
        {(keyword "/?:@-._~!$'()* ,;") "/?:@-._~!$'()* ,;=="}))
 
+(deftest t-path-params
```
ohpauleez: Two quick things:
 * We should add 'negative' tests here to ensure that paths that don't require decoding aren't modified in unintended ways
 * We should add another test that exercises this functionality using the routes tress.

service/src/io/pedestal/http.clj
```diff
@@ -255,6 +255,7 @@
                                           (csrf/anti-forgery enable-csrf)])
                true (conj (middlewares/content-type {:mime-types ext-mime-types}))
                true (conj route/query-params)
+               true (conj route/path-params)
```
ohpauleez: Since this is a breaking change and has a slight performance impact, we should setup an option to turn this off -- by default it will be turned on.

ohpauleez:
This looks good to me.  Have you previously signed the electronic CLA [linked in the Contributing guide](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md#contributors-agreement)?

joakimmohn:
Great. Yes, I have signed the CCA.

ohpauleez:
Awesome, thanks!

route/src/io/pedestal/http/route.clj
```diff
@@ -121,12 +121,21 @@
                     key
                     (.append b c)))))))))
 
+(defn parse-param-map [m]
+  (persistent! (reduce-kv (fn [acc k v] (assoc! acc k (decode-query-part v))) (transient {}) m)))
```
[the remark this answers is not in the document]
ohpauleez: I'm going to address that in master with another change.  I think the service/interceptor option should take an interceptor and not a boolean -- ie: Provide me an interceptor to decode path-params and by default I'll URL-decode them; nil turns it off.  Much like your `+` idea, I came up with another desired way I might want to decode path-params which was slightly different than URL decoding.

Issue #560, opened 2018-05-09
SSE dispatch loop abort without proper clean up


kotokz:
# Description
The current SSE dispatch loop not able to handle exception which thrown by send-event function. 
The loop simply abort without closing the event channel and calling on-client-disconnect call back.

This issue heavy impact application which depends on the on-client-disconnect function.
we fixed it by putting loop under a try-finally block to handle the resource cleanup

## Expected Behavior
the event-channel, response-channel should be closed
on-client-disconnect

## Actual Behavior
only response channel can be closed.

## Steps to reproduce
trigger a exception which only can be caught by "send-event" function. 
for example, trigger a channel buffer full scenario: 
[async-dispatch-3] io.pedestal.http.sse -- {:msg "exception sending event", :throwable #error {
 :cause "Assert failed: No more than 1024 pending puts are allowed on a single channel. Consider using a windowed buffer.\n(< (.size puts) impl/MAX-QUEUE-SIZE)"
 :via
 [{:type java.lang.AssertionError
   :message "Assert failed: No more than 1024 pending puts are allowed on a single channel. Consider using a windowed buffer.\n(< (.size puts) impl/MAX-QUEUE-SIZE)"
   :at [clojure.core.async.impl.channels.ManyToManyChannel put_BANG_ "channels.clj" 152]}]
 :trace
 [[clojure.core.async.impl.channels.ManyToManyChannel put_BANG_ "channels.clj" 152]
  [clojure.core.async$put_BANG_ invokeStatic "async.clj" 165]
  [clojure.core.async$put_BANG_ invoke "async.clj" 158]
  [io.pedestal.http.sse$send_event invokeStatic "sse.clj" 93]
  [io.pedestal.http.sse$send_event invoke "sse.clj" 81]
  [io.pedestal.http.sse$start_dispatch_loop$fn__8544$state_machine__5275__auto____8557$fn__8560 invoke "sse.clj" 133]
  [io.pedestal.http.sse$start_dispatch_loop$fn__8544$state_machine__5275__auto____8557 invoke "sse.clj" 133]
  [clojure.core.async.impl.ioc_macros$run_state_machine invokeStatic "ioc_macros.clj" 973]
  [clojure.core.async.impl.ioc_macros$run_state_machine invoke "ioc_macros.clj" 972]
  [clojure.core.async.impl.ioc_macros$run_state_machine_wrapped invokeStatic "ioc_macros.clj" 977]
[...]

Issue #557, opened 2018-05-03
`nil` hosts should be handled by Pedestal, not containers


ohpauleez:
If a service map is missing the `:http/host` key or otherwise has a `nil` value, the resolution of the host binding is determined by the container implementation.

I'm proposing a breaking change.  When Pedestal processes the service map and [prepares the server options map](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http.clj#L339), we should set a default `:host` of `127.0.0.1`.  In the spirit of being security-focused, we should prevent folks from accidentally exposing services, and in the spirit of consistency, we should unify behavior across various containers when possible.

To signal to developers that the host should be purposefully set, we should uncomment the [host setting in service template](https://github.com/pedestal/pedestal/blob/master/service-template/src/leiningen/new/pedestal_service/service.clj#L70) and add appropriate docs/comments instructing the developer how/why to set new host values.

ohpauleez:
This fix is now on master.

Issue #555, opened 2018-05-01
Illegal reflective access warning on loading jetty classes


rafaeldff:
# Description
Opening a Java 9 JVM in a project with `[io.pedestal/pedestal.jetty "0.5.3"]` will show _Illegal reflective access_ warnings. This apparently was fixed in jetty upstream (https://github.com/eclipse/jetty.project/issues/1262).

## Expected Behavior
Expected not to see any warnings

## Actual Behavior
See warning such as
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.eclipse.jetty.util.BufferUtil (file:/home/rafael/.m2/repository/org/eclipse/jetty/jetty-util/9.4.0.v20161208/jetty-util-9.4.0.v20161208.jar) to field java.nio.MappedByteBuffer.fd
WARNING: Please consider reporting this to the maintainers of org.eclipse.jetty.util.BufferUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

## Steps to reproduce
Open a REPL with Java 9 on a project depending on `[io.pedestal/pedestal.jetty "0.5.3"]`, then require pedestal.jetty - `(require '[io.pedestal.http.jetty])`. You will see the warnings as above.

# Environment

## Operating System (including version).
Linux rffnix 4.15.17 #1-NixOS SMP Thu Apr 12 10:31:21 UTC 2018 x86_64 GNU/Linux

## Your current Leiningen or Boot version (`lein --version` or `boot --version`)
Leinigen 2.8.1
java 9.0.4
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)

## Pedestal version
```
[io.pedestal/pedestal.service "0.5.3" :exclusions [org.slf4j/slf4j-nop]]
[io.pedestal/pedestal.service-tools "0.5.3"] 
[io.pedestal/pedestal.jetty "0.5.3"]
```

ohpauleez:
As you noted, this is fixed in upstream Jetty (the project made a big effort to address all compatibility issues).  Given Java 9's EOL (March 2018), most JDK compatibility issues will be focused at Java 10, in preparation for Java 11 (which has LTS).

It's unclear how the Java and Clojure communities will react to the "Support Cliffs" in the newly established release cycle.

ohpauleez:
[...]

Issue #554, opened 2018-05-01
Correctly handle nil values in :headers map when using io.pedestal.test/response-for


kennyjwilli:
# Description
Calling `response-for` with `nil` as a header value will result in an exception being logged as an ERROR. The error message is not thrown in the REPL. Additionally, an incorrect response map is returned.

## Expected Behavior
`nil` values in the `:headers` map are handled gracefully with no exception thrown. The correct response map is returned.

## Actual Behavior
[This exception](https://pastebin.com/P6A0Pm9S) is thrown in the logs when called with a nil header value (see below steps). Also an incorrect response map is returned: 

```clojure
{:status nil, :body "", :headers nil}
```

## Steps to reproduce

```clojure
(io.pedestal.test/response-for
  service-fn
  :post "/my-route"
  :headers {"cookie" nil})
=> {:status nil, :body "", :headers nil}
```

This exception is caused by the elements array passed to `String/join` being `null` [here](https://github.com/pedestal/pedestal/blob/e9230c3a49b7036b195267c7c24b0c4881c941b5/service/src/io/pedestal/http/request.clj#L27). `(enumeration-seq (.getHeaders servlet-req key))` is null.

## Pedestal version
`[io.pedestal/pedestal.service "0.5.3"]`

ohpauleez:
Hi @kennyjwilli - thanks for reporting the issue!

Is your expectation that `nil` should work because of your experience elsewhere (ie: another web library, testing tool, or live traffic against a Pedestal web application)?  A nil value doesn't conform to the [Ring Spec](https://github.com/ring-clojure/ring/blob/master/SPEC).

What is your expectation about how `nil` header values should be gracefully handled?  How should a `nil` value be interrupted?

Given that it doesn't conform to the spec, I'd expect an exception to be thrown, but in this case, was the exception not informative?

kennyjwilli:
Given that it does not conform to the Ring Spec, I'd also expect an exception to be thrown. The exception linked above is rather obtuse, however; it's not at all clear that the issue is because my `:headers` map contains a `nil` value. I think a better approach would be to throw as soon as we know the `:headers` map does not pass the spec, rather than throwing when `String/join` is passed `nil`.

mtnygard:
@ohpauleez I asked @kennyjwilli to submit an issue when I saw his questions on Slack. He got a stack trace that started in stylobate. I don't think we should ask users to understand stylobate or depend on its implementation.

ohpauleez:
[...]

PR #565, opened 2018-05-24
Bump dependencies; Various bugfixes

ohpauleez:
This PR addresses a handful of issues and stale PRs:

 * Deps cleaned up #549 
 * All Clojure and Clojure contrib deps bumped up -- Clojure 1.9.0 is standard across the project
 * Jetty deps bumped up, fixing the JDK9 issue in the process (#555 #552)
 * Tomcat deps bumped up, fixing an initialization bug that was lingering
 * Undertow/Immutant deps bumped
 * SSE metric increment/decrement correctly accounted for in all branches #548 
 * SSE safely runs the handling loop in a try/finally block, to ensure shutdown and cleanup always happens #560
 * nil `::host` in service map defaults to localhost now (rather than being container specific) #557
 * Content-negotiation response map now conforms to Ring spec #540
 * `io.pedestal.log` now has MDC capabilities, supported with a new protocol #551
 * `response-for` now catches nil header values early, with a helpful exception #554 
 * Updated the AWS Lambda sample to use the new aws module.


ohpauleez:
Regarding the tests for the MDC stuff - since it's an interface and a protocol, there's substantial to test.  I could (and will) certainly write a test for the macros

PR #566, opened 2018-05-25
Small bug fix plus doc additions.

ddeaguiar:
Fixed the default `set-mdc` implementation and expanded on the mdc-specific docs.

### DIFF
--- a/log/src/io/pedestal/log.clj
+++ b/log/src/io/pedestal/log.clj
@@ -168,7 +168,7 @@
     t)
   (-set-mdc [t m]
     (when m
-      (.setContextMap t ^Map t))
+      (.setContextMap t ^Map m))
     t)
 
   nil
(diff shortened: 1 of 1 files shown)


ddeaguiar:
@ohpauleez, I held off on adding another `with-context-*` variant since I wanted to avoid introducing yet another way to work with the mdc context.

log/src/io/pedestal/log.clj
```diff
@@ -333,21 +333,36 @@
...
+  This map also includes all options that were passed into `with-context`."
   {})
 
 (def mdc-context-key "io.pedestal")
 
 (defmacro with-context
   "Given a map of keys/values/options and a body,
-  Set the map into the SLF4J MDC via the *mdc-context* binding.
+  Set the map into the MDC via the *mdc-context* binding.
+  The MDC used defaults to SLF4J MDC unless the `io.pedestal.log/mdc`
```
ohpauleez: Might want to change this to `:io.pedestal.log/mdc` so readers know it's a keyword

ddeaguiar:
@ohpauleez updated. Thanks!

ohpauleez:
:shipit:

Issue #578, opened 2018-06-19
Support Jetty context path


terjesb:
# Description

I've been looking into migrating some Pedestal services from WildFly/Immutant to standalone Jetty. When run under WildFly, the deployment gets its context path from the container. When deploying the same route table to Jetty, it's now in the root context. This makes it inconvenient to target both containers during a transition. There may also be other use-cases for deploying the same app to Jetty under multiple/different context paths. Or even running multiple Jetty apps behind a single load balancer, and having them under different contexts/prefixes instead of all in the root context on different ports.

One option is to (programmatically) update the routes, i.e. have separate route tables for the different containers.
Another option would be to use the same route table, but enable Jetty to use a context similar to WildFly.

I've looked into the latter and have a patch for this with passing tests:
* add support for a :context-path in Jetty's :container-options
* change url-for-routes to support existing :context parameter (it had an if branch for this already, but I didn't get it to work and it doesn't appear to be tested)
* change servlet-interceptor to add :context-path to the request again, to support dynamic url-for via :request :context-path and the same if branch as above. (Ref also our previous discussion in issue #439.) Tried to do this with an 'external' interceptor, but then the default-interceptors and routing plumbing seem to adds the url-for function to the context _before_ the :context-path is available. Not sure how to prepend an interceptor to the chain. (default-interceptors both constructs the vector and updates service-map, perhaps these two operations could be separated later.)

I think the changes are only breaking for anyone using url-for and its :context parameter, but I believe this did not work as documented.

(Note that url-for probably doesn't work for WildFly/Immutant right now either, ref the previous discussion. This patch doesn't fix this, but possibly makes it easier to do. OTOH, if url-for isn't wanted/required for Jetty either, adding only :context-path seems trivial.)

Preparing a PR.

## Expected Behavior

Pedestal could support Jetty's context-path and make it more convenient to transition between WildFly and Jetty deployment, or support the same app under different contexts.

## Actual Behavior

Pedestal does not support Jetty's context-path.

## Steps to reproduce

# Environment

## Pedestal version

0.5.3

ohpauleez:
Thanks for reporting the issue and pulling together a PR @terjesb!  I agree that these changes should be made and the PR looks good.  I'll be pulling together a few changes for 0.5.4 today -- if I touch `url-for`, I'll just pull your changes in, otherwise, I'll just merge your PR when I'm setting up my PR.

ohpauleez:
[...]

Issue #439, opened 2016-06-03
:context-path not in 0.5.0?


terjesb:
I have an app deployed to WildFly 10/Immutant 2 that works on some 0.4.2-SNAPSHOT and stops working on 0.5.0. It relies on req :context-path from the container.

I noticed that io.pedestal.http.request.map now has these commented out:

```
;(assoc! :servlet-context (.getServletContext ^ServletConfig servlet))
;(assoc! :context-path (.getContextPath servlet-req))
;(assoc! :servlet-path (.getServletPath servlet-req))
```

What's the recommended migration here?

ohpauleez:
Hi @terjesb, thanks for the report (and for trying 0.5.0 out before the announcement)!  I was hoping a few people would try it out to sort out these issues before it got announced more publicly.

With the creation on Chain Providers, the request got more generic (dropping the servlet specific pieces), exposing only a single function for servlet-related request creation.  Those pieces were tentatively going to be removed (because you could always get to them from `:servlet`, `:servlet-request`, and `:servlet-response` -- Just saving space and time per request for the common case.

Are you looking at the `:context-path` directly in your application?  If so, you should replace that code with something like `(some-> (:servlet-request response) (.getContextPath))`?  Is the application failing [in routing](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/route.clj#L180-L194) -- maybe because you're using `url-for` or the sequential router or passing your own function for routing?

terjesb:
Thanks, @ohpauleez. So great seeing Pedestal move forward! I'm currently using :context-path via a locally patched version of propan/geheimtur (auth for Pedestal; looking forward to seeing what bits Arachne will use/provide in that area) in order to get proper login-uri routing on WildFly.

So shouldn't be a problem to add your replacement there. Haven't used url-for yet, but hopefully those pieces will also somehow work on servlet containers going forward? (Btw, also posted a table-routes question to the mailing list.)

ohpauleez:
As part of the 0.5.1 effort, we're gll loooing to start pulling community projects (and community-powered documentation) more into the fold, which should help situations like this.  `url-for` should work as expected unless a developer is manually doing something interesting with the context's request.  Nonetheless, keep me posted as things pop up, and I'll ensure things move forward nicely for you!

ohpauleez:
Report back here once everything works out as expected and we'll close out the issue.

terjesb:
[...]

PR #579, opened 2018-06-19
Support Jetty context-path in container-options and url-for

terjesb:
Proposed patch for issue #578.

In route.clj, I think maybe the (concat context-path-parts path-parts) could replace that cond.

# Description

I've been looking into migrating some Pedestal services from WildFly/Immutant to standalone Jetty. When run under WildFly, the deployment gets its context path from the container. When deploying the same route table to Jetty, it's now in the root context. This makes it inconvenient to target both containers during a transition. There may also be other use-cases for deploying the same app to Jetty under multiple/different context paths. Or even running multiple Jetty apps behind a single load balancer, and having them under different contexts/prefixes instead of all in the root context on different ports.

One option is to (programmatically) update the routes, i.e. have separate route tables for the different containers.
Another option would be to use the same route table, but enable Jetty to use a context similar to WildFly.

I've looked into the latter and have a patch for this with passing tests:
* add support for a :context-path in Jetty's :container-options
* change url-for-routes to support existing :context parameter (it had an if branch for this already, but I didn't get it to work and it doesn't appear to be tested)
* change servlet-interceptor to add :context-path to the request again, to support dynamic url-for via :request :context-path and the same if branch as above. (Ref also our previous discussion in issue #439.) Tried to do this with an 'external' interceptor, but then the default-interceptors and routing plumbing seem to adds the url-for function to the context _before_ the :context-path is available. Not sure how to prepend an interceptor to the chain. (default-interceptors both constructs the vector and updates service-map, perhaps these two operations could be separated later.)

I think the changes are only breaking for anyone using url-for and its :context parameter, but I believe this did not work as documented.

(Note that url-for probably doesn't work for WildFly/Immutant right now either, ref the previous discussion. This patch doesn't fix this, but possibly makes it easier to do. OTOH, if url-for isn't wanted/required for Jetty either, adding only :context-path seems trivial.)

Preparing a PR.

## Expected Behavior

Pedestal could support Jetty's context-path and make it more convenient to transition between WildFly and Jetty deployment, or support the same app under different contexts.

## Actual Behavior

Pedestal does not support Jetty's context-path.

## Steps to reproduce

# Environment

## Pedestal version

0.5.3

### DIFF
--- a/jetty/src/io/pedestal/http/jetty.clj
+++ b/jetty/src/io/pedestal/http/jetty.clj
@@ -147,13 +147,16 @@
          port :port
          {:keys [ssl? ssl-port
                  h2? h2c? connection-factory-fns
-                 context-configurator configurator max-threads daemon? reuse-addr?]
+                 context-configurator context-path configurator max-threads daemon? reuse-addr?]
           :or {configurator identity
+               context-path "/"
                max-threads (max 50 (needed-pool-size))
                h2c? true
                reuse-addr? true}} :container-options} options
         thread-pool (QueuedThreadPool. ^Integer max-threads)
         server (Server. thread-pool)
+        _ (when-not (clojure.string/starts-with? context-path "/")
+            (throw (IllegalArgumentException. "context-path must begin with a '/'")))
         _ (when (and h2? (not ssl-port))
             (throw (IllegalArgumentException. "SSL must be enabled to use HTTP/2. Please set an ssl port and appropriate *store setups")))
         _ (when (and (nil? port) (not (or ssl? ssl-port h2?)))
@@ -186,7 +189,7 @@
                           (.setReuseAddress reuse-addr?)
                           (.setPort ssl-port)
                           (.setHost host)))
-        context (doto (ServletContextHandler. server "/")
+        context (doto (ServletContextHandler. server context-path)
                   (.addServlet (ServletHolder. ^javax.servlet.Servlet servlet) "/*"))]
     (when daemon?
       (.setDaemon thread-pool true))
@@ -243,4 +246,3 @@
   ;; :trust-password - the password to the truststore
   ;; :client-auth  - SSL client certificate authenticate, may be set to :need,
   ;;                 :want or :none (defaults to :none)"
-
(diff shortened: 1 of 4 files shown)


service/src/io/pedestal/http/impl/servlet_interceptor.clj
```diff
@@ -202,6 +202,7 @@
              ;                                                 :servlet-request servlet-request
              ;                                                 :servlet-response servlet-response})
              :async? servlet-async?)
+      (assoc-in [:request :context-path] (some-> servlet-request (.getContextPath)))
```
ohpauleez: I appreciate the defensive coding here and normally I'd require it, but this function is in the hot path for handling even simple requests.  If `servlet-request` is nil, I'd rather have the server blow up and return 500s.  No need to make this change, I'll make it in another branch/PR

jetty/src/io/pedestal/http/jetty.clj
```diff
@@ -147,13 +147,16 @@
...
-                 context-configurator configurator max-threads daemon? reuse-addr?]
+                 context-configurator context-path configurator max-threads daemon? reuse-addr?]
           :or {configurator identity
+               context-path "/"
                max-threads (max 50 (needed-pool-size))
                h2c? true
                reuse-addr? true}} :container-options} options
         thread-pool (QueuedThreadPool. ^Integer max-threads)
         server (Server. thread-pool)
+        _ (when-not (clojure.string/starts-with? context-path "/")
```
ohpauleez: I'm going to add a `:require` for clojure.string in another branch, even though it's technically not needed.

service/test/io/pedestal/http/jetty_context_path_test.clj
```diff
@@ -0,0 +1,79 @@
+;; This test file is a port of the official Ring jetty-adapter test
+;; that works with Pedestal interceptors. The original version is
+;; here:
+;;
+;; https://github.com/ring-clojure/ring/blob/master/ring-jetty-adapter/test/ring/adapter/test/jetty.clj
+
```
ohpauleez: I'm going to relocate this under `http/jetty_test.clj`.

Issue #582, opened 2018-07-03
Easier handling of Character Encoding / Content-Length


orestis:
# Description

First of all, thanks for all the hard work for creating and maintaining Pedestal. I'm having tons of fun :)

I'm a relatively newcomer to Clojure and the JVM, and after trying to setup a simple HTTP JSON API on top of Pedestal, I was surprised to find that this is not there out of the box, nor there are samples to show how to handle them — and googling doesn't seem to help either. I fixed my immediate use case, but I'm filing an issue after discussion on Slack.

Imagine an API that returns a JSON object. I expect the final HTTP response to be encoded in UTF-8, and a correct Content-Length header to be added. Following the [Hello world with content types](http://pedestal.io/guides/hello-world-content-types) I can see how I can convert a hash map to JSON, using the library of my choice. However, it's not at all clear from the documentation when an where the `String` that most JSON libraries provide gets turned into bytes, and how I can tap into this.

I've managed to come up with this on my own:

```
(def coerce-to-json
  {:name ::coerce-to-json
   :leave
   (fn [context]
     (let [response         (get context :response)
           body             (get response :body)
           coerced-body     (.getBytes (json/generate-string body) StandardCharsets/UTF_8)
           updated-headers  (assoc (get response :headers {})
                                   "Content-Type" "application/json"
                                    "Content-Length" (str (count coerced-body)))
           updated-response (assoc response
                                   :headers updated-headers
                                   :body    coerced-body)]
       (assoc context :response updated-response)))})
```

But I would really expect that there would be a default interceptor that does both of those out of the box — having an out of the box Content Length working is hugely important. I realise this might be a Ring issue, since I couldn't find *anything* on this. I've tested this in Jetty but I'm not sure how it would look like under other servers.

## Expected Behavior

An Example on how to serve JSON using a built-in interceptor that took care of the low-level details would be nice :) 

## Pedestal version

0.5.4

mk:
Hi @orestis,

There's the json-body interceptor in pedestal that does just that:
https://github.com/pedestal/pedestal/blob/09dd88c4ce7f89c7fbb7a398077eb970b3785d2d/service/src/io/pedestal/http.clj#L115-L127

To use it, you set the response body to a collection type and make sure the json-body interceptor comes afterwards in your interceptor chain. You can also enqueue it dynamically like so:

```clojure
(require '[io.pedestal.interceptor.chain :as chain]
         '[io.pedestal.http :as http])

(def my-api-interceptor
[...]

PR #583, opened 2018-07-03
Custom Response Interceptor sample

ohpauleez:
This is just another sample that shows how to tightly control the response type, format, content, headers using an interceptor.  The example uses a simple JSON API for motivation, related to #582

### DIFF
--- a/samples/json-api/.gitignore
+++ b/samples/json-api/.gitignore
@@ -1,0 +1,38 @@
+# Project related
+
+# Java related
+pom.xml
+pom.xml.asc
+*jar
+*.class
+
+# Leiningen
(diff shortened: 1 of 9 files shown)


orestis:
Hi Paul, thanks for getting on top of this, it's going to be very helpful. As I mentioned in the mailing list thread, I don't think we can get away from calling `.getBytes` as `count` will count unicode code points and not converted bytes:

```
(count "Orestis") ;; => 7
(count "Ορέστης") ;; => 7
(count (.getBytes "Orestis")) ;; => 7
(count (.getBytes "Ορέστης")) ;; => 14
```

PR #584, opened 2018-07-25
Change index? into :index-files?

tcoenraad:
As https://ring-clojure.github.io/ring/ring.util.response.html#var-file-response shows that the option for rendering index files is called `:index-files?` and not `:index?` this should be reflected in the documentation and default options for `io.pedestal.http.ring-middlewares/fast-resource`.

I think my changes can be seen as
> sufficiently trivial

and as a result I have written no tests.

### DIFF
--- a/service/src/io/pedestal/http/ring_middlewares.clj
+++ b/service/src/io/pedestal/http/ring_middlewares.clj
@@ -149,11 +149,11 @@
   This interceptor will return async responses for large files (files larger than the HTTP Buffer)
   If your container doesn't recognize FileChannel response bodies, this interceptor will cause errors
   Supports a map of options:
-  :index? - If path is a directory, will attempt to find an 'index.*' file to serve. Defaults to true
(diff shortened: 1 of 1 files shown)


ohpauleez:
Hi @tcoenraad - thanks for the PR!

Since `fast-resource` is a Pedestal-only interceptor, argument name compatibility with other Ring middleware isn't a top priority, but certainly eases adoption/switching.

In briefly looking at the code, I think the docstring for `fast-resource` is incorrect, in that it doesn't support the index file argument and requires a direct resource file reference.

I'm going to close this pull-request -- we should either update the code to make it conform to the docs (if it doesn't) or update the docstring to more accurately reflect the controls of the interceptor.  Additionally, changing default key names in general is a breaking change and usually requires backwards compatibility to be added (with tests that show both old and new version work as expected).  This would potentially be an exception, since that key doesn't appear to be used anyway.

PR #585, opened 2018-08-07
Docstrings for create-server, start, stop.

cloojure:
Title says it all.


ohpauleez:
This is a good idea and we should merge something very similar to this.

We should update the format/wording to be consistent with the other docstrings ("Given ... returns .... Notes: .... See: ...").  Additionally the docstrings should be expanded a bit to explain things like the expected arity of `:start-fn` and where it gets added to the service map, as well as what part of the system typically calls things like `start`

ddeaguiar:
Thanks for this. Refined and resolved in #620.

Issue #581, opened 2018-06-25
Terminator check after async interceptor appears to be broken


hlship:
# Description

If an async interceptor adds a valid :response key to the context, the interceptor chain should terminate, but doesn't.

I thought I remembered this as a known bug, but was unable to find it.

## Expected Behavior

After an async interceptor executors (conveys the updated context to the returned channel) a check for a terminator (e.g., a valid :response key) should occur; when the :response key is present and valid, the interceptor chain should terminate, and `leave` handlers should start to execute.

## Actual Behavior

```
29.156 DEBUG io.pedestal.interceptor.chain - {:in begin, :execution-id 6, :line 314}
29.156 DEBUG io.pedestal.interceptor.chain - {:in process-all, :handling :enter, :execution-id 6, :line 154}
29.157 DEBUG io.pedestal.interceptor.chain - {:interceptor :io.pedestal.http.impl.servlet-interceptor/terminator-injector, :stage :enter, :execution-id 6, :fn #object[io.pedestal.interceptor.helpers$before$fn__17403 0x7603b6ab "io.pedestal.interceptor.helpers$before$fn__17403@7603b6ab"], :line 50}
29.157 DEBUG io.pedestal.interceptor.chain - {:interceptor :io.pedestal.http.impl.servlet-interceptor/stylobate, :stage :enter, :execution-id 6, :fn #object[io.pedestal.http.impl.servlet_interceptor$enter_stylobate 0x5f6d5c89 "io.pedestal.http.impl.servlet_interceptor$enter_stylobate@5f6d5c89"], :line 50}
29.157 DEBUG io.pedestal.interceptor.chain - {:interceptor :io.pedestal.http/log-request, :stage :enter, :execution-id 6, :fn #object[io.pedestal.interceptor.helpers$on_request$fn__17459 0x673589d5 "io.pedestal.interceptor.helpers$on_request$fn__17459@673589d5"], :line 50}
29.157 DEBUG io.pedestal.interceptor.chain - {:interceptor :io.pedestal.http.route/query-params, :stage :enter, :execution-id 6, :fn #object[io.pedestal.http.route$fn__18325 0x1aa5c507 "io.pedestal.http.route$fn__18325@1aa5c507"], :line 50}
[...]

PR #589, opened 2018-09-01
Fix async termination

ddeaguiar:
Addresses the issue described in #581. The terminators are checked in the initial loop binding of `process-all-with-binding`.


ddeaguiar:
@ohpauleez addressed the issue as you described and added a test.

PR #591, opened 2018-09-07
Fix bad import

puredanger:
As of core.specs.alpha 0.2.44, the spec for import has been tightened and will now fail if you use `(:import (package.class))`, which has always been wrong (and actually silently did nothing). Should actually be `(:import (package class))`.


ohpauleez:
This should not have been merged.

The import was intended to be the the full package+class.  The mistake was wrapping it in parens.  @ddeaguiar Please post a PR for the fix ASAP

PR #594, opened 2018-09-18
Fix calling arity to match documentation

bltavares:
It is documented on the `trace` interceptor that the calling convention
for the `span-processor` would be a single-arg function, receiving the
context.

When implementing an alternative span-processor, an `arity exception`
arose, as it is being called with two-args.

This commit changes the calling function to receive a single arg.
This should not break both of the provided processors, xray and
default-processor.


ddeaguiar:
@ohpauleez this PR LGTM and a signed CA exists.

PR #597, opened 2018-10-15
Use ByteArrayOutputStream as buffer instead of outputting directly to…

tcoenraad:
… HttpOutput

The current behavior of `transit-body-interceptor` directly writes transit encoded data to HttpOutput. This makes it work very slow. By using the transit writer as proposed in the docs of transit-clj (https://github.com/cognitect/transit-clj#usage), thus by using a in-memory buffer, the encoding time goes down to very acceptable levels.

In our own project, the time it took to finish a request with about 100 entities went from about 7 seconds to about 0,25 seconds.

### DIFF
--- a/service/src/io/pedestal/http.clj
+++ b/service/src/io/pedestal/http.clj
@@ -30,7 +30,8 @@
             [cognitect.transit :as transit]
             [io.pedestal.log :as log])
   (:import (java.io OutputStreamWriter
-                    OutputStream)))
+                    OutputStream
+                    ByteArrayOutputStream)))
 
 ;; edn and json response formats
 
@@ -152,8 +153,10 @@
           (-> response
               (ring-response/content-type default-content-type)
               (assoc :body (fn [^OutputStream output-stream]
-                             (transit/write
-                              (transit/writer output-stream transit-format transit-opts) body)
+                             (let [out (ByteArrayOutputStream. 4096)
+                                   writer (transit/writer out transit-format transit-opts)]
+                               (transit/write writer body)
+                               (.writeTo out output-stream))
                              (.flush output-stream))))
           response))))))
 


jaimeagudo:
Sounds a no brainer, feel free to merge, just curious, have you benchmarked with different buffer sizes (changing `4096`) ?

tcoenraad:
It did not have any real impact setting the buffer size bigger or smaller, so I took the arbitrary number from the docs on `transit-clj`.

jaimeagudo:
Results on production may vary, it might worth to measure it there at some point too :)

ddeaguiar:
@tcoenraad thanks for the PR. In the future please create an issue for such change requests. @jaimeagudo, in the future please leave the merge decisions to the project maintainers.

ddeaguiar:
@tcoenraad please review the [Contributing](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md#contributors-agreement) guidelines. You will need to sign and submit a Contributor's Agreement before this change could be merged if approved.

tcoenraad:
If I remember correctly, I have already signed and submitted a Contributor's Agreement for https://github.com/pedestal/pedestal/pull/584, so I thought it was still applicable. But if that has expired, I'm glad to do so still.

Nevertheless, may I ask whether this change will be picked up after all?

mtnygard:
@tcoenraad The change makes sense to me. We're likely to merge it, but give us a bit to track down your CA. (We need to check in with the person who keeps the records.)

ohpauleez:
Thanks everyone for the contributions and trying to make Pedestal better!

I'm dropping some comments here as folks dig deeper into reviewing this:

 * What was the cause of slowness in going directly to HttpOutput?  Is that slowness a deeper issue that should be addressed?  I have a few assumptions, but we should investigate that to better understand the impact of this change and if there is a better solution.
 * I think the fixed, in-memory buffer should be an option that is passed into the interceptor at creation time.  This would allow those who want the stream-to-stream streaming behavior to stay in place, without introducing the in-memory fixed buffer.  It would also allow those to tune the size of the buffer that best fits their application.

In general, we tend to discourage PRs without opening issues first.  Issues give a chance for the wider community to comment and shape the proposed solution.

Thanks!

PR #600, opened 2018-11-07
refactor: simpler nil-check

teresy:
a simple refactor to use `nil?` check where possible

### DIFF
--- a/service/src/io/pedestal/http/request/zerocopy.clj
+++ b/service/src/io/pedestal/http/request/zerocopy.clj
@@ -62,7 +62,7 @@
     (dissoc (request/realized this) k))
 
   clojure.lang.Counted
-  (count [this] (unchecked-add (if (= base-request nil) 0 (count request/ring-dispatch))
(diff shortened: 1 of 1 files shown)


ddeaguiar:
@ohpauleez pointed out that  there may be places where `=` is used vs `nil?` for performance purposes. This change may not have an effect here but it's something to keep in mind in other areas of the codebase.

teresy:
@ddeaguiar thanks for the note. FWIW, this suggestion is implemented in [kibit](https://github.com/jonase/kibit) -- I'd be happy to learn of ways that we can detect cases where `= .. nil` is better than `nil?` for this pattern so that kibit could take it into account.

ddeaguiar:
@teresy I suspect it's going to be case-specific and I doubt a static analysis tool like kibit can properly classify the nuanced cases. In this case, the zerocopy implementation is performance sensitive. Paul has done an excellent job keeping performance considerations in mind throughout the Pedestal implementation and provided me with additional context. 

TBH, this note was as much for my benefit (future self) as for anyone else.

PR #605, opened 2018-12-18
Update readme to use 0.5.5

MageMasher:


ddeaguiar:
Thanks for catching that!

PR #607, opened 2019-01-25
Update CodeStar sample to the latest versions/details of AWS Services

ohpauleez:
This PR updates the CodeStar sample with the updates related to AWS Service updates.

This includes a new version of CodeBuild and CodePipeline, new rules around how proxies work on API Gateway, Canary support on APIGW, updates to Lambda runtimes/limits, and CloudFormation updates.

### DIFF
--- a/samples/aws-codestar-lambda/config/logback.xml
+++ b/samples/aws-codestar-lambda/config/logback.xml
@@ -14,7 +14,7 @@
 
     <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
       <!-- rollover daily -->
-      <fileNamePattern>logs/pub-consumer-%d{yyyy-MM-dd}.%i.log</fileNamePattern>
+      <fileNamePattern>logs/pedestal-lambda-%d{yyyy-MM-dd}.%i.log</fileNamePattern>
       <!-- or whenever the file size reaches 64 MB -->
       <maxFileSize>64 MB</maxFileSize>
     </rollingPolicy>


samples/aws-codestar-lambda/src/pedestallambda/service.clj
```diff
@@ -2,9 +2,83 @@
...
+  This is achieved by first looking to see if the key already exists, and if not,
+  adding the information as reported in the Content-Type header.
+  This is necessary because `body-params` expects `:content-type`, but some
+  chain providers (like APIGW) do the bare minimum when making the request map/object."
+  (interceptor/interceptor
+    {:name  ::pre-body-params
+     :enter (fn [ctx]
+              (if (get-in ctx [:request :content-type])
+                ctx
+                (assoc-in ctx [:request :content-type] (or (get-in ctx [:request :headers "content-type"])
```
ddeaguiar: Based on the doc-string I gather that the request object may not be a clojure map. Will it be `associative?`? Otherwise this won't work, will it?

samples/aws-codestar-lambda/src/pedestallambda/service.clj
```diff
@@ -2,9 +2,83 @@
...
+;; `reduce-kv` doesn't drop back to generic `java.util.Map` support
+;; like `reduce` and other sequence-oriented functions in clojure.core.
+;; This is a known issue in Clojure: https://dev.clojure.org/jira/browse/CLJ-1762
+;;
+;; This is needed because Pedestal uses `reduce-kv` when processing Map-like
+;; requests and APIGW uses java.util.Maps.
+;;
+;; THIS SHOULD BE REMOVED ONCE CLOJURE IS PATCHED!
+;; ----
+(extend-protocol clojure.core.protocols/IKVReduce
```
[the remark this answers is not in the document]
ohpauleez: The various `params` in the Request are all Maps.  Most of Pedestal's params processing (including all the `keywordize` param stuff, and transitively `body-params`) uses reduce-kv, and will blow up unless this patch is in place.  Arguably, this should have already been fixed in Clojure.

samples/aws-codestar-lambda/src/pedestallambda/service.clj
ohpauleez: Every one of Pedestal's request-map implementations supports Associative (including the one-off/standalone implementation used in the Lambda/APIGW provider).  This works as expected.  The "object/map" is a reference that it could be satisfy the Associative interface but be some other object underneath (other than a persistent map, as seen in the zerocopy request).

All of Clojure's core abstractions are interfaces (maps, seqs, associative), so `assoc` never implies a concrete data type, just something that implements the appropriate interfaces.

samples/aws-codestar-lambda/src/pedestallambda/service.clj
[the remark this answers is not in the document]
ddeaguiar: Got it, thanks.

samples/aws-codestar-lambda/src/pedestallambda/service.clj
ddeaguiar: 👍

PR #609, opened 2019-02-18
Update gzip sample with comments about GzipHander default behavior.

souenzzo:


ohpauleez:
Thanks!  It's a useful suggestion.  I just need someone on the Pedestal team to verify the suggestions work and we can get this merged.

ddeaguiar:
@ohpauleez, I had discussed this with @souenzzo in the #pedestal slack room. I'll take a closer look and merge if ok soonish.

samples/servlet-filters-gzip/src/gzip/service.clj
```diff
@@ -61,6 +61,11 @@
               ::http/container-options {:context-configurator (fn [c]
                                                                 (let [gzip-handler (GzipHandler.)]
                                                                   (.setGzipHandler c gzip-handler)
+                                                                  ;; by default, "GzipHandler" only operates on "GET" 's.
+                                                                  ;; It is possible to add other methods via:
+                                                                  ;; (.addIncludedMethods gzip-handler (into-array ["GET" "POST"]))
```
ddeaguiar: Can we remove `"GET"`? Including it here is unnecessary although it does no harm to do so.

ohpauleez: Unnecessary, but it's nice to see it explicitly listed.

ddeaguiar: Ok.

PR #614, opened 2019-03-15
'nil' :path-params are not allowed.

### LINKED ISSUE #602: The url-for :strict-path-params? option should be more strict.
The `url-for` `:strict-path-params?` option was introduced to address issue #572, however, it still allows `nil` values to be passed in as `path-params`.

Given the routes `#{["/foo/bar/:baz" :get [`foo]]}`, the following invocation should throw `(url-for ::foo :strict-path-params? true :path-params {:baz nil})`.

ddeaguiar:
This PR resolves #602. We should not allow `nil` values for path-params when using `url-for`. It also compares sets of path param keys since `keys` does not guarantee order.

### DIFF
--- a/route/src/io/pedestal/http/route.clj
+++ b/route/src/io/pedestal/http/route.clj
@@ -258,7 +258,7 @@
                             (set (seq (:path-params route))) ;; match the params from the route?  `seq` is used to handle cases where no `path-params` are required
                             )
                       ;; nils are not allowed.
-                      (reduce-kv #(if (nil? %3) (reduced true)  %1) false path-params)))
+                      (reduce-kv #(if (nil? %3) (reduced true)  false) nil path-params)))
             (throw (ex-info "Attempted to create a URL with `url-for`, but missing required :path-params - :strict-path-params was set to true.
                             Either include all path-params, or if your URL actually contains ':' in the path, set :strict-path-params to false in the options"
                             {:path-parts path-parts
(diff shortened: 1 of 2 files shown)


ohpauleez:
Thanks Dan for pulling this together.  I think we might want to consider doing the inner bit with `reduce-kv`/`reduced` to prevent seq'ing over the map multiple times and keep the fail-fast characteristic.

ddeaguiar:
@ohpauleez thanks for the feedback and sorry for the delay. I've changed the impl accordingly.

route/src/io/pedestal/http/route.clj
```diff
@@ -253,9 +253,12 @@
...
-                     (not= (keys path-params) ;; Do the params passed in...
-                           (seq (:path-params route)) ;; match the params from the route?  `seq` is used to handle cases where no `path-params` are required
-                           ))
+                     (or
+                      (not= (set (keys path-params)) ;; Do the params passed in...
+                            (set (seq (:path-params route))) ;; match the params from the route?  `seq` is used to handle cases where no `path-params` are required
+                            )
+                      ;; nils are not allowed.
+                      (reduce-kv #(if (nil? %3) (reduced true)  false) nil path-params)))
             (throw (ex-info "Attempted to create a URL with `url-for`, but missing required :path-params - :strict-path-params was set to true.
```
solussd: Should this error message also state that nils are not allowed?

PR #618, opened 2019-05-18
Remove two-arg call to activate

theJohnnyBrown:
The two-arity version is [removed in version 0.33](https://github.com/opentracing/opentracing-java/commit/5ed6a47975ed9a19c8de3bcbe27a9385a14de479#diff-1fba031275dd47ef915364ad0aa5c267) and the one-arity call is present in 0.31

### DIFF
--- a/log/src/io/pedestal/log.clj
+++ b/log/src/io/pedestal/log.clj
@@ -839,7 +839,7 @@
                initial-tags)
        (.start ^Tracer$SpanBuilder builder))))
   (-activate-span [t span]
-    (.activate (.scopeManager t) ^Span span false)
+    (.activate (.scopeManager t) ^Span span)
     span)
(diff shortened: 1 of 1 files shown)


theJohnnyBrown:
@ohpauleez anything else you need to see here?

ohpauleez:
Thanks for checking in!
I think we're all set here.  We're just waiting until all dependencies are updated and we can get some additional tests around this change.

ddeaguiar:
@theJohnnyBrown, thanks for creating this PR! Can you submit a signed [Cognitect Contributor Agreement](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md#contributors-agreement)?

ddeaguiar:
CA completed.

PR #621, opened 2019-06-13
Do not log Broken pipe exception

tonsky:
When using with Immutant, Pedestal logs Broken pipe errors quite often. As far as I understand, those are nothing to be afraid of, just client silently closing TCP connection while other side is not informed about it and tries to write. This is a part of normal operation and definitely should not dominate your logs:

Screenshot 2019-06-05 at 18 26 58

### DIFF
--- a/service/src/io/pedestal/http/impl/servlet_interceptor.clj
+++ b/service/src/io/pedestal/http/impl/servlet_interceptor.clj
@@ -29,7 +29,7 @@
             [ring.util.response :as ring-response])
   (:import (javax.servlet Servlet ServletRequest ServletConfig)
            (javax.servlet.http HttpServletRequest HttpServletResponse)
-           (java.io OutputStreamWriter OutputStream EOFException)
+           (java.io OutputStreamWriter OutputStream EOFException IOException)
            (java.nio.channels ReadableByteChannel)
            (java.nio ByteBuffer)))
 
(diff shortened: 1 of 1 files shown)


ddeaguiar:
Hi @tonsky, I really appreciate your interest in making Pedestal better! While I appreciate the intent to reduce log noise, I don't think such a specific change belongs in the `error-ring-response` fn. As per its doc string, this function is intended to to a catch all exception handler and it's preferable for service implementations to catch and handle exceptions in the way that best suites the service.

That being said, although possible, there is no easy way for a service implementor to handle and exception as responses are streaming. As per the `http-interceptor-service-fn` [implementation](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/impl/servlet_interceptor.clj#L374-L376), errors thrown by the `ring-response` [fn](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/impl/servlet_interceptor.clj#L288) are immediately handled by the error stage impl of the stylobate [interceptor](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/impl/servlet_interceptor.clj#L286).

Ideally a service implementor would have the ability to provide a service-specific error handling interceptor to handle errors such as this. The bottom line is that I think a more general solution is needed. I'm going to create an issue about this and close this PR.

ddeaguiar:
Issue @621 created.

PR #625, opened 2019-07-15
Fix header key name for OpenTracing span

DjebbZ:


ohpauleez:
Thanks for cleaning up my sloppy doc/comment typos, they're difficult for me to spot :)

DjebbZ:
Arf, please don't close the PR I was typing the description of the issue.

DjebbZ:
Hello, I'd like to exchange this small pull-request for some support regarding implementing OpenTracing for our pedestal web app.

Using `pedestal 0.5.5` and `io.jaegertracing/jaeger-client "0.35.5"` I managed to add OpenTracing to my application, I can spawn the Jaeger UI and see the traces, whether it's the trace automatically created for each request by pedestal or the ones I manually add. So far so good.

The problem I have is that we want to generate the Trace ID not from our Pedestal web app but from our NginX in the front. This part works fine, but our pedestal app is not able to extract the Trace ID from the HTTP headers, so we can't connect the dots the see the flow of the request. We need because down the line we have more pedestal API apps for which we must pass the Trace ID in the headers.

Looking closely at the problem, it seems that :

1. The headers is [supposed to be named `uber-trace-id`](https://github.com/jaegertracing/documentation/blob/62b43b32f4fa4bf9c317fb66818dab7c1d9cfcf4/content/docs/1.9/client-libraries.md#propagation-format), not `uber-id` (hence my PR). It seems confirmed by the NginX module we installed that does generate the proper `uber-trace-id` header (and not `uber-id`).
2. The code that does the extraction `io.pedestal.interceptor.trace/default-span-resolver` doesn't seem to work. I noticed it's different from the one in the AWS xray tracing : in `io.pedestal.log.aws.xray/span-resolver`, the resolver for HTTP headers looks like :

``` clojure
;; Is there an X-Ray Trace ID in the headers?
(when-let [header-str (or (get-in context [:request :headers trace-header-lower])
                          (get-in context [:request :headers TraceHeader/HEADER_KEY]))]
  (let [^TraceHeader trace-header (TraceHeader/fromString ^String header-str)
        ^TraceID trace-id         (.getRootTraceId trace-header)
        ^String parent-id         (.getParentId trace-header)]
    ;; Defend against the case where you're cycling back in on yourself,
    ;; within the same thread.
    (if-let [current-entity (try
                              (.getTraceEntity recorder)
                              (catch Exception e nil))]
      (.beginSubsegment recorder ^String operation-name)
      (.beginSegment recorder
                     ^String operation-name
                     ^TraceID trace-id
                     ^String parent-id))))
```

It seems to try reading from the header directly and doesn't use the TextMapExtractAdapter from OpenTracing. Are you sure the one in `io.pedestal.interceptor.trace/default-span-resolver` works ?

I'd love help on this point, because it's the only remaining blocker that prevents us from deploying our pedestal app with OpenTracing.

PS: I promise that if you help me solved the problem, I'll do a PR to update the *outdated* tracing sample (I have a working version :) )

DjebbZ:
After a few trials, it seems that the code used to extract the SpanContext (and the Trace ID) from the HTTP headers always returns nil =>

``` clojure
(.extract ^Tracer log/default-tracer
          Format$Builtin/HTTP_HEADERS
          (TextMapExtractAdapter. {"uber-trace-id" "8fbaf02bafb708d4079586ac617d177b"}))
;; nil
```

This could explain why it doesn't work.

DjebbZ:
Even when using a 64-bit long trace id in the header (my example was way too long), it sill returns `nil`.

PR #627, opened 2019-08-02
Fix typo ('interceptors' for possessive singular)

SherSpock:
This is a dangling bit of business left over from pedestal/pedestal-docs/issues/125


SherSpock:
yw!

PR #628, opened 2019-08-08
Throw informative ex-info with some context instead of NPE if scheme nil

codeasone:
Here's a [minimal reproduction](https://github.com/codeasone/url-for-npe) in case it's helpful in clarifying the scenario where this bit me.

### DIFF
--- a/route/src/io/pedestal/http/route.clj
+++ b/route/src/io/pedestal/http/route.clj
@@ -281,7 +281,10 @@
         port-mismatch   (not= port   request-port)]
     (str
      (when (or absolute? scheme-mismatch host-mismatch port-mismatch)
-       (str (when (or absolute? scheme-mismatch) (str (name scheme) \:))
+       (str (when (or absolute? scheme-mismatch)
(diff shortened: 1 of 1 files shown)


ddeaguiar:
@codeasone, thanks for your interest in contributing to Pedestal! The issue seems limited to tests based on a particular usage. In general, once your app starts depending on more complex routing needs, it's preferable to create a `url-for` test helper which can be used to generate the routes for tests. In your sample repo, you can do something like this:

```
(def routes #{["/dummy" :get `show :route-name :dummy]})

(def url-for
  (route/url-for-routes (route/expand-routes routes) :scheme :http :host "localhost" :port 8080))

(:body (response-for (::http/service-fn running)
                       :get (url-for :dummy)
                       :headers {"Accept" "text/html"}))
```
I'm going to close the PR for now. If you feel strongly about it, feel free to open an issue to discuss further.

codeasone:
That's fine @ddeaguiar I understand. Thanks for getting back to me.

Issue #626, opened 2019-08-02
X-Ray fails on tag-span when span created with :initial-tags using a Clojure map


ddeaguiar:
# Description
When creating a span using a clojure map as the value of `:initial-tags` and subsequently calling `tag-span`, X-Ray throws an `java.lang.UnsupportedOperationException` exception.

```
{:via
 [{:type java.lang.UnsupportedOperationException,
   :at [clojure.lang.APersistentMap put "APersistentMap.java" 380]}],
 :trace
 [[clojure.lang.APersistentMap put "APersistentMap.java" 380]
  [rentals.service$eval19545 invokeStatic "NO_SOURCE_FILE" 43]
  [rentals.service$eval19545 invoke "NO_SOURCE_FILE" 43]
  [clojure.lang.Compiler eval "Compiler.java" 7177]
  [clojure.lang.Compiler eval "Compiler.java" 7132]
  [clojure.core$eval invokeStatic "core.clj" 3214]
  [clojure.core.server$prepl$fn__8941 invoke "server.clj" 232]
  [clojure.core.server$prepl invokeStatic "server.clj" 228]
  [clojure.core.server$prepl doInvoke "server.clj" 191]
  [clojure.lang.RestFn invoke "RestFn.java" 425]
  [cognitect.rebl$repl invokeStatic "rebl.clj" 120]
  [cognitect.rebl$repl invoke "rebl.clj" 94]
  [cognitect.rebl$_main invokeStatic "rebl.clj" 123]
  [cognitect.rebl$_main invoke "rebl.clj" 122]
  [clojure.lang.AFn applyToHelper "AFn.java" 152]
  [clojure.lang.AFn applyTo "AFn.java" 144]
  [clojure.lang.Var applyTo "Var.java" 705]
  [clojure.core$apply invokeStatic "core.clj" 665]
  [clojure.core.server$accept_connection invokeStatic "server.clj" 73]
  [clojure.core.server$start_server$fn__8879$fn__8880$fn__8882
   invoke
   "server.clj"
   117]
  [clojure.lang.AFn run "AFn.java" 22]
  [java.lang.Thread run "Thread.java" 748]],
 :phase :execution}
```

[Here's](https://github.com/aws/aws-xray-sdk-java/blob/v2.2.1/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java#L450) the offending line in the X-Ray SDK.

## Expected Behavior
[...]

PR #632, opened 2019-10-11
Extract OpenTracing context from headers

vielmath:
I had some trouble working with `jaeger` and `pedestal 0.5.7`
The `tracing-interceptor` wasn't able to extract the span context from the request headers.

> WARN  i.j.i.PropagationRegistry$ExceptionCatchingExtractorDecorator - Error when extracting SpanContext from carrier. Handling gracefully.
> java.lang.ClassCastException: class io.opentracing.propagation.TextMapExtractAdapter cannot be cast to class io.opentracing.propagation.TextMap (io.opentracing.propagation.TextMapExtractAdapter and io.opentracing.propagation.TextMap are in unnamed module of loader 'app')

Using a `TextMapAdapter` instead of a `TextMapExtractAdapter` seems to fix the issue.
I added a simple test to make sure something was extracted from the `uber-trace-id` header.

I also updated the `tracing` and `tracing-interceptor` samples to use the last deps.
The issue can be reproduced running the updated `tracing-interceptor` sample.

### DIFF
--- a/interceptor/src/io/pedestal/interceptor/trace.clj
+++ b/interceptor/src/io/pedestal/interceptor/trace.clj
@@ -5,7 +5,12 @@
   (:import (io.opentracing Tracer
                            SpanContext)
            (io.opentracing.propagation Format$Builtin
-                                       TextMapExtractAdapter)))
+                                       TextMapAdapter)))
+
+(defn headers->span-context [^Tracer tracer headers]
+  (.extract tracer
+            Format$Builtin/HTTP_HEADERS
+            (TextMapAdapter. headers)))
 
 (defn default-span-resolver
   ([context]
@@ -26,9 +31,8 @@
                                         (.getAttribute servlet-request "TracingFilter.activeSpanContext"))]
              (log/span operation-name ^SpanContext span-context))
            ;; Is there an OpenTracing span in the headers? (header key is "uber-trace-id")
-           (when-let [span-context (.extract ^Tracer log/default-tracer
-                                             Format$Builtin/HTTP_HEADERS
-                                             (TextMapExtractAdapter. (get-in context [:request :headers] {})))]
+           (when-let [span-context (headers->span-context log/default-tracer
+                                                          (get-in context [:request :headers] {}))]
              (log/span operation-name ^SpanContext span-context))
            ;; Otherwise, create a new span
            (log/span operation-name))
--- a/samples/tracing-interceptor/project.clj
+++ b/samples/tracing-interceptor/project.clj
@@ -1,24 +1,24 @@
-(defproject tracing-interceptor "0.5.5"
+(defproject tracing-interceptor "0.5.7"
   :description "FIXME: write description"
   :url "http://example.com/FIXME"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies [[org.clojure/clojure "1.10.0"]
-                 [io.pedestal/pedestal.service "0.5.5"]
+  :dependencies [[org.clojure/clojure "1.10.1"]
+                 [io.pedestal/pedestal.service "0.5.7"]
 
                  ;; Remove this line and uncomment one of the next lines to
                  ;; use Immutant or Tomcat instead of Jetty:
-                 [io.pedestal/pedestal.jetty "0.5.5"]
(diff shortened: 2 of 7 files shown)


ohpauleez:
Thanks so much for pulling this together!

Once I get up to speed on the changes in the latest Jaeger release and the changes to the continued-spans (via headers) in OpenTracing, we can get these changes merged.

ohpauleez:
Unusually that the code in question was causing an issue.  `Extract` should be the read-only aspect of union'd interface, `TextMap`.  I think we'll have to dig into the OpenTracing code to figure out what's going on here -- it could be that TextMap is being used as the interface now internally to simplify integration points ala https://medium.com/opentracing/opentracing-java-0-32-released-a3c9b6af08e4

vielmath:
I found this solution in the jaeger-client-java tests:
https://github.com/jaegertracing/jaeger-client-java/blob/0b2d27c8105da8f98121063ea68e09dd2272a49f/jaeger-core/src/test/java/io/jaegertracing/internal/AdhocHeadersTest.java#L84

ddeaguiar:
> Thanks so much for pulling this together!
> 
> Once we get up to speed on the changes in the latest Jaeger release and the changes to the continued-spans (via headers) in OpenTracing, we can get these changes merged.
> 
> Dan, if you want to confirm all these changes:
> 
> * Changes looked fine to me, but review until you're content
> * Ensure there isn't an alternative method of continued spans we want to expose
> * Work through the idea if we want make the `headers->continued-span` a protocol so we can also extend it to X-Ray (if that's of interest to folks or clients).

Thanks for the PR @vielmath and thanks @ohpauleez for the initial review! @vielmath, have you filled out a [Contributor's Agreement](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md#contributors-agreement)?

I'll dive deeper into this but the OpenTracingApi introduced various breaking changes with the `0.32` and this one was missed.  The class `TextMapExtractAdapter` [no longer](https://github.com/opentracing/opentracing-java/blob/release-0.32.0/opentracing-api/src/main/java/io/opentracing/propagation/TextMapExtractAdapter.java#L29) implements the `TextMap` interface which is one of the interfaces [expected](https://github.com/opentracing/opentracing-java/blob/release-0.32.0/opentracing-api/src/main/java/io/opentracing/Tracer.java#L112-L113) to be supported by `Tracer/extract`.

Overall, the changes look ok to me although I'm not initially keen on the jaeger test dependency for pedestal.service but haven't considered alternatives. 

It could be that these changes are too specific and that the initial implementation was, similarly (see below).

I'm not sure yet if `headers->continued-span` should be a protocol method (I feel something like this should implemented symmetrically). Inject/Extract are part of the OpenTracingApi and the intent is to use Carriers to do so. My instincts tell me that allowing the specification of a carrier would be more useful. That being said, we _already_ support specifying a custom span resolver when constructing the tracing interceptor which could be leveraged as the workaround for this issue until a new Pedestal release lands.

Anyways, thanks folks and I'll spend more time digging in!

ddeaguiar:
> @vielmath implemented this fix to avoid us having to implement a custom interceptor that actually retrieves the uber-trace-id from the headers.

@DjebbZ, in the meantime you can implement a custom span resolver, no need for a separate interceptor! It's one of the [options](https://github.com/pedestal/pedestal/blob/c2ff3810c3b2aa445b0501e2909de320b2143418/interceptor/src/io/pedestal/interceptor/trace.clj#L86) you can pass to `tracing-interceptor`.

DjebbZ:
Yes we saw that, thanks for the notice.

vielmath:
> Thanks for the PR @vielmath and thanks @ohpauleez for the initial review! @vielmath, have you filled out a [Contributor's Agreement](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md#contributors-agreement)?
> 

I just did, sorry.

ddeaguiar:
After some thought, I don't think there's a need for an additional protocol since we already provide an option for specifying a span resolver.

DjebbZ:
Great job @vielmath !

@ddeaguiar do you know when you will cut a release including this fix ?

ddeaguiar:
> Great job @vielmath !
> 
> @ddeaguiar do you know when you will cut a release including this fix ?

Not sure yet. I'd like to at least get a fix in for https://github.com/pedestal/pedestal/issues/626 before cutting a release.

PR #633, opened 2019-10-18
Catch tracer creation exceptions.

### LINKED ISSUE #624: Missing catch in try? Or useless try?
# Description

In https://github.com/pedestal/pedestal/blob/master/log/src/io/pedestal/log.clj#L978, the code that tries to resolve the OpenTracing's GlobalTracer is wrapped in a `try` without a `catch`/`finally`. Unless I'm missing something, either somebody forgot the `catch`/`finally`, or somebody removed the `catch`/`finally` and forgot to remove the `try`.

## Pedestal version

0.5.5+

ddeaguiar:
Default to the `NoopTracer` when an exception is raised during tracer creation. Resolves #624.

### DIFF
--- a/log/src/io/pedestal/log.clj
+++ b/log/src/io/pedestal/log.clj
@@ -970,7 +970,12 @@
   The value of the setting should be a namespaced symbol
   that resolves to a 0-arity function or nil.
   That function should return something that satisfies the TracerOrigin protocol.
-  If no function is found, the GlobalTracer will default to the NoopTracer and `GlobalTracer/isRegistered` will be false."
+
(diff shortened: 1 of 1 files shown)


ddeaguiar:
@ohpauleez, the `default-tracer` doc-string was ambiguous in regards to this and there was no `catch`. This change does change behavior. I'm ok with removing the try entirely if bubbling the exception up was the original intent.

ohpauleez:
The floating `try` is suspicious to me.  Typically I tend to let configuration errors blow up rather than "silently" continuing forward.

I would just drop the try and retain the existing behavior.

ddeaguiar:
Ok, will close the PR and kill the try.

Issue #634, opened 2019-10-22
io.pedestal.test/test-servlet-response reifys HttpServletResponse but not sendError() methods


hlship:
# Description

Code which de-references the context's :servlet-response key and invokes .sendError fails by throwing java.lang.AbstractMethodError

## Expected Behavior

The error code or error code and plaintext body should become the response 

## Actual Behavior

Exception thrown

## Steps to reproduce

(-> context :servlet-response (.sendError 401))

# Environment

OpenJDK 1.8

## Operating System (including version).

OS X 10.12.6

## Your current Leiningen or Boot version (`lein --version` or `boot --version`)

Leiningen 2.7.1 on Java 1.8.0_212 OpenJDK 64-Bit Server VM

## Pedestal version

0.5.3

ohpauleez:
Hi Howard!

Thanks for reporting the issue.  Is there any additional context you can provide about ways you're interacting with the Servlet directly?
Also, what would be your general expectations about calling `.sendError` on the reified servlet -- what information gets stored (if any), where should it be stored (if there is data), and how are you validations within the test shaped where you make this call?

hlship:
I think our use case is a bit ugly ... from what I can tell, the goal is to send the response immediately and bypass the :leave stage of any pending interceptors (the same code removes :io.pedestal.interceptor.chain/stack and :io.pedestal.interceptor.chain/queue from the queue).  

... and looking at the code, we normally only call that in one specific case.  

Ideally, I'd expect to see test-servlet-response return a Ring response with just the :status code provided to .setError.

ohpauleez:
Excellent, thanks for the extra bit of context.  Your expectation of result matches mine as well -- should be a quick change to make.  Thanks!

PR #636, opened 2019-11-21
Adds test support for HttpServletResponse SendError.

ddeaguiar:
Support HttpServletResponse's `sendError` method with and without a message.

Addresses #634 .


ohpauleez:
Yep, this is how I was planning on doing it.  Thanks!

PR #639, opened 2020-01-10
Return StringBuffer on getRequestURL

thumbnail:
When calling `(.getRequestURL req)` on a request created by `response-for` the following error popped up:
```
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.StringBuffer
```

Which is logical because `javax.servlet.http.HttpServletRequest#getRequestUrl` expects to return a `StringBuffer` instead of a `String`.

---

Happy to add a tests, but I'd need a suggestion on the approach, the fn is private and creating an interceptor for testing the request seems non-ideal.

### DIFF
--- a/service/src/io/pedestal/test.clj
+++ b/service/src/io/pedestal/test.clj
@@ -106,7 +106,7 @@
         (getMethod [this] (-> verb
                             name
                             cstr/upper-case))
-        (getRequestURL [this] url)
+        (getRequestURL [this] (StringBuffer. url))
         (getServerPort [this] port)
         (getServerName [this] host)
         (getRemoteAddr [this] "127.0.0.1")


thumbnail:
Hey @ddeaguiar 👋🏻,

 anything i can do to get this moving?

ddeaguiar:
> Hey @ddeaguiar 👋🏻,
> 
> anything i can do to get this moving?

Hi @thumbnail, thanks for the PR. I'll take a look at it in the next couple of days. In the meantime, have you filled out a [Contributor Agreement](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md)?

thumbnail:
Thanks for bringing the CA to my attention, it has since been submitted 👍

PR #647, opened 2020-03-05
support async handler function

lucywang000:
To use the async handler feature of pedestal, the handler function must be wrapped in a full interceptor, for instance with the  `interceptor.helpers/defbefore` macro:

```clojure
(defbefore handle-login
  [context]
  (async/go
    (async/<! (async/timeout 1000))
    (assoc context :response {:status 200 :body "hello"}))
```

This patch make it able to be simplified to:

```clojure
(defn hello
  [request]
  (async/go
    (async/<! (async/timeout 1000))
    {:status 200 :body "hello"}))
```

### DIFF
--- a/interceptor/src/io/pedestal/interceptor.clj
+++ b/interceptor/src/io/pedestal/interceptor.clj
@@ -12,7 +12,8 @@
 
 (ns io.pedestal.interceptor
   "Public API for creating interceptors, and various utility fns for
-  common interceptor creation patterns.")
+  common interceptor creation patterns."
+  (:require [clojure.core.async :as async]))
 
 (defrecord Interceptor [name enter leave error])
 
@@ -24,6 +25,7 @@
   (-interceptor [t] "Given a value, produce an Interceptor Record."))
 
 (declare interceptor)
+(declare channel?)
 (extend-protocol IntoInterceptor
 
   clojure.lang.IPersistentMap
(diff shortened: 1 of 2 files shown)


ohpauleez:
Hi @lucywang000 - thanks for using Pedestal and having interest in contributing!

Pedestal supports two approaches for async behavior.  An interceptor can "go async" and transfer the handling to another thread (this is done at the interceptor level) or a response body can be a WriteableAsyncBody, triggering the container's async handling capabilities and enabling async handling all the way down to the wire (ie: through Pedestal to the I/O).  [There exists](https://github.com/pedestal/pedestal/blob/master/service/src/io/pedestal/http/impl/servlet_interceptor.clj#L100) a backwards-compatibility to allow response bodies to be core.async channels (much like you've done in your PR) even though this is discouraged.

It was a specific design choice to make async interceptors more evident in the code via
`(def my-async-interceptor {:enter (fn [ctx] ...)})`.

We also tried to keep the core abstractions and APIs (eg: interceptor.clj) as free of dependencies as possible and as free from conditional logic as possible (favoring polymorphic extension and composition).

I'm happy to see more evidence about where these decisions cause problems or hardships, or where alternatives open up other possibilities that are currently not within reach.  The community should weigh in too!
I'm going to close this PR, but we can open an issue or move this to the mailing list.

lucywang000:
Hi @ohpauleez Thanks for the reply. 

> ... a response body can be a WriteableAsyncBody

That's a good feature, but only the body is allowed to be an async channel. This means the application has to set the status code, headers, etc., before going async.


> It was a specific design choice to make async interceptors more evident in the code via
(def my-async-interceptor {:enter (fn [ctx] ...)}).

May I ask why? 

If you take a look at the two examples in my PR description again, I'd argue it's way more counter-intuitive to force people to write a function that takes a context map but the only usage of the context is to get the `:request` key out of it.

In pedestal handler functions are wrapped into interceptors under the hood, which provides a conceptual integrity. However most of the time (90%+ imo) application developers are writing code with handlers, not interceptors, which justifies the importance of the `handlers` and the benefit it would yield if we facilitate the way to write async handlers out of the box.


> We also tried to keep the core abstractions and APIs (eg: interceptor.clj) as free of dependencies as possible and as free from conditional logic as possible

Well the interceptor/chain.clj ns depends on core.async anyway and it's full of conditional logic :P .

PR #653, opened 2020-05-09
Remove deref call in routes function in service-template

bpringe:
I may be misunderstanding something but from what I can tell using the `service/routes` symbol directly inside the function will always get the current value of `service/routes` when the `server/routes` function is called, so dereferencing the var is unnecessary. This change works for me - changes to the routes are picked up without a server restart.


ddeaguiar:
Hi @bpringe, the usage of `deref` is by intent so that routes are reloaded during development.

bpringe:
Yes, I realize that's the intent, but the `deref` is unnecessary/redundant since it's a function referring to var. The latest value for routes will always be used. Given the change in this PR, it still works as intended, as far as I can tell and have tested.

PR #654, opened 2020-05-13
Allow overriding the Jetty thread pool

tobias:
This allows consumers to provide a customized thread pool to Jetty via
:container-options.

We are currently using this in [a fork](https://clojars.org/io.clubhouse.forks/pedestal.jetty) to inject an [`InstrumentedQueuedThreadPool`](https://metrics.dropwizard.io/3.1.0/manual/jetty/).

Let me know if you would prefer a different approach to overriding this.

### DIFF
--- a/jetty/src/io/pedestal/http/jetty.clj
+++ b/jetty/src/io/pedestal/http/jetty.clj
@@ -144,6 +144,16 @@
   ([connectors acceptors selectors]
    (* (Math/round ^Double (+ acceptors selectors)) connectors)))
 
+(defn- thread-pool
+  "Returns a thread pool for the Jetty server. Can be overridden
+  with [:container-options :thread-pool] in options. max-threads is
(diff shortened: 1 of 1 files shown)


ohpauleez:
Looks totally fine to me.  On line 244, I would include that the option ignores `:max-threads`

tobias:
Thanks @ohpauleez. I just force-pushed a new commit with your suggestion and the error that was caught by CI.

ddeaguiar:
Hi @tobias, thanks for contributing to Pedestal! Have you signed a [Contributor's Agreement](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md#contributors-agreement)?

tobias:
@ddeaguiar I have now, thanks for the reminder!

tobias:
Thanks for merging @ddeaguiar. Any idea of when you will be doing the next release?

PR #656, opened 2020-05-15
Adds support for jetty websocket flow control.

### LINKED ISSUE #497: Websocket handlers should handle flow control/backpressure asynchronously
# Description

Context jetty/websocket: 
At first glance it seems that messages sent to websockets are fire & forget, the callback on the jetty side that receives the ACK (http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/websocket/api/WriteCallback.html) is not used, also see http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/websocket/api/RemoteEndpoint.html#sendBytes-java.nio.ByteBuffer-org.eclipse.jetty.websocket.api.WriteCallback- . That means that a slow consumer could cause the handler to drop messages and exceptions to be raised (not to mention it'd cause the channel buffer to fill for nothing instead of signaling the need to exert backpressure upstream). Also the handler would never know about potential write errors.

source in question: https://github.com/pedestal/pedestal/blob/master/jetty/src/io/pedestal/http/jetty/websockets.clj#L22-L51

ws-send could return a promise-chan that'd signal ACK from the receiver and we'd be just one <! away of a proper solution.

ddeaguiar:
This PR adds a new websocket connect function, `start-ws-connection-with-fc-support`, along with a protocol `WebSocketSendAsync` which is internally non-blocking and optionally supports flow control. 

Additionally, while the intended behavior of `start-ws-connection` was left intact, additional error handling was added to ensure the go loop does not terminate when invalid input is provided.

The use of either _start_ fn is optional and service implementers are free to provide their own and integrate directly with the websocket Session if neither impl fits their needs.

Resolves #497.

### DIFF
--- a/jetty/src/io/pedestal/http/jetty/websockets.clj
+++ b/jetty/src/io/pedestal/http/jetty/websockets.clj
@@ -1,20 +1,22 @@
 (ns io.pedestal.http.jetty.websockets
-  (:require [clojure.core.async :as async :refer [go-loop]])
+  (:require [clojure.core.async :as async :refer [go-loop]]
+            [io.pedestal.log :as log])
   (:import (java.nio ByteBuffer)
            (org.eclipse.jetty.servlet ServletContextHandler ServletHolder)
            (org.eclipse.jetty.websocket.servlet WebSocketCreator
-                                                WebSocketServlet
-                                                WebSocketServletFactory)
+                                                WebSocketServlet)
            (org.eclipse.jetty.websocket.api Session
                                             WebSocketListener
                                             WebSocketConnectionListener
-                                            RemoteEndpoint)))
+                                            RemoteEndpoint
+                                            WriteCallback)))
 
 
 ;; This is a protocol used to extend the capabilities on messages are
 ;; marshalled on send
 (defprotocol WebSocketSend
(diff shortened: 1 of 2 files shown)


ddeaguiar:
@ohpauleez interested in your thoughts on this one.

jetty/src/io/pedestal/http/jetty/websockets.clj
```diff
@@ -42,14 +43,84 @@
...
+      p-chan))
+
+  ByteBuffer
+  (ws-send-async [msg ^RemoteEndpoint remote-endpoint]
+    (let [p-chan (async/promise-chan)]
+      (.sendBytes remote-endpoint msg (reify WriteCallback
+                                        (writeFailed [_ ex]
+                                          (async/put! p-chan ex))
+                                        (writeSuccess [_]
+                                          (async/put! p-chan :success))))
```
lucywang000, line 83: The reify for  `WriteCallback ` looks 100% the same as the one for String, so I think it could be refactored out (e.g. a deftype?)

ddeaguiar: Yep.

PR #657, opened 2020-05-15
fast-resource pass on context unmodified when request is not GET or HEAD

Whoops:
Trying to diagnose why I was not getting a 404 when POSTing to routes that didn't exist, I finally figured out that the `fast-resource` interceptor returns `nil` when the request method is not GET or HEAD. This changes it to instead return `ctx` unmodified.

### DIFF
--- a/service/src/io/pedestal/http/ring_middlewares.clj
+++ b/service/src/io/pedestal/http/ring_middlewares.clj
@@ -190,7 +190,8 @@
                                                                           (into-array OpenOption [StandardOpenOption/READ])))))]
                        (if response
                          (assoc ctx :response response)
-                         ctx)))))}))))
+                         ctx))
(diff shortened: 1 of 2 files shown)


ddeaguiar:
Hi @Whoops, thanks for troubleshooting this. Have you filled out a [Contributor's Agreement](https://github.com/pedestal/pedestal/blob/master/CONTRIBUTING.md#contributors-agreement)?

Whoops:
I hadn't, but I have now!

ddeaguiar:
FYI the build is failing due to SSL handshake errors since the PR builds are apparently running against Java 11 and the test keystore is using an algorithm which is not valid with TSL 1.3. An updated test keystore file has been pushed to master.

PR #668, opened 2020-09-28
update hello world example

AndreaCrotti:
Copy hello world from the docs (since defroutes is deprecated)

### DIFF
--- a/samples/hello-world/src/hello_world/service.clj
+++ b/samples/hello-world/src/hello_world/service.clj
@@ -12,21 +12,17 @@
 
 (ns hello-world.service
   (:require [io.pedestal.http :as http]
-            [io.pedestal.http.route :as route]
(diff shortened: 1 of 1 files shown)


ohpauleez:
Hi @AndreaCrotti - Thanks so much for jumping in to make Pedestal better!

Some small tweaks to consider:
 * I would normally not call `expand-routes` as part of the `def` for routes.  Pedestal will call `expand-routes` when it processes the service-map.  I try to keep all the top-level defs and inputs into Pedestal as end-user data representations.
 * I would usually syntax quote the route handler in the route table to avoid the explicit `:route-name`.  Some folks prefer the explicit route-name, especially when heavily using `url-for`.
 * Both of these suggestions can be seen on the main Pedestal site - http://pedestal.io/

Thanks again!

PR #691, opened 2021-10-18
Update the sample projects to the latest release

claj:
I did go through all the sample projects to upgrade them. I did not reproduce everything in every README-file (I have not tried out capstan, not deployed things to AWS etc).

I had problems with some functionality, mostly `http2`, without conscrypt, which are known bugs. I did not put a lot of effort into finding a decently old JVM to get these working.

I have updated all the occurences of the alpn-agent, event in out commented `project.clj` files, and of course updated all the other deps to reasonable versions.

**some notes:**

`aws-codestar-lambda` - I have not tested to deploy this to AWS. There are many reflection warnings, mainly related to boxing. The backported `reduce-kv` on `java.util.Map` is not removed since we use latest stable Clojure 1.10.3.
`fast-pedestal`- disabled several deprecated JVM options meant to optimize the speed. Works with JDK 17 but I did not do any speed measurments.
`helloworld-metrics` - it is not clear to me if the reported metrics are complete.
`http2` fails because it is missing ALPN handlers for JDK 17, but also for JVMs like 8.0.302-zulu.
`http2-conscrypt`- works using 8.0.302-zulu
`tracing` and `tracing-interceptor`- had to make exclusion in the updated deps. Works nicely.
`tracing-interceptor-aws` - tried with local xray-daemon and I can see the traces on AWS.
`war-example`- I have not tried running the generated `.war`-file.

Please let me know if it something should be changed in the apporach, like a test matrix for different jvm:s (probably using sdkman) and/or some test scripts based on the instructions in the READMEs.


bonkydog:
OK, I've updated this branch and tested that all the samples still start and respond with a 200.
Had to update a jetty-servlet dependency.

bonkydog (approved):
Looks good. Thanks!

Issue #696, opened 2021-12-15
Jetty websocket start-ws-connection (indirectly) calls a blocking operation inside a go block


souenzzo:
# Description

I found a blocking call inside a go-loop.

As described in the docstring

https://github.com/clojure/core.async/blob/v1.5.648/src/main/clojure/clojure/core/async.clj#L441

"go blocks should not (either directly or indirectly) perform operations that may block indefinitely"

In pedestal, we are calling indirectly a operation that describes itself as a `blocking` operation (and throws IOException)

Stacktrace:
`go-loop`
https://github.com/pedestal/pedestal/blob/0.5.9/jetty/src/io/pedestal/http/jetty/websockets.clj#L48
`ws-send`
https://github.com/pedestal/pedestal/blob/0.5.9/jetty/src/io/pedestal/http/jetty/websockets.clj#L51
`ws-send@String - .sendString`
https://github.com/pedestal/pedestal/blob/0.5.9/jetty/src/io/pedestal/http/jetty/websockets.clj#L25
`.sendString`
https://github.com/eclipse/jetty.project/blob/jetty-9.4.35.v20201120/jetty-websocket/websocket-api/src/main/java/org/eclipse/jetty/websocket/api/RemoteEndpoint.java#L108 (edited) 

## Steps to reproduce

# Environment

## Pedestal version

0.5.9

ddeaguiar:
Thanks for reporting @souenzzo. There is an async alternative for ws-send. Perhaps we can leverage that instead.

souenzzo:
Can we  add a descriptive docstring warning about this issue and recommending the async alternative?

ddeaguiar:
We can add a warning to the doc-string but I'm not sure about recommending the async alternative just yet since the mechanics are unclear to me atm.

PR #706, opened 2022-01-27
#696: Remove sync io from go blocks

aredington:
- Deprecate the WebSocketSend protocol
- Remove default implementations of WebSocketSend protocol
- Rewrite `start-ws-connection` in terms of `ws-send-async`
- Conveys exceptions on write in band over the promise channel as per `start-ws-connection-with-fc-support`

### DIFF
--- a/jetty/src/io/pedestal/http/jetty/websockets.clj
+++ b/jetty/src/io/pedestal/http/jetty/websockets.clj
@@ -28,34 +28,6 @@
   (ws-send [msg ^RemoteEndpoint remote-endpoint]
     (.sendBytes remote-endpoint msg)))
 
-(defn start-ws-connection
-  "Given a function of two arguments
-  (the Jetty WebSocket Session and its paired core.async 'send' channel),
-  and optionall a buffer-or-n for the 'send' channel,
-  return a function that can be used as an OnConnect handler.
-
-  Notes:
-   - You can control the entire WebSocket Session per client with the
-  session object.
-   - If you close the `send` channel, Pedestal will close the WS connection."
-  ([on-connect-fn]
-   (start-ws-connection on-connect-fn 10))
-  ([on-connect-fn send-buffer-or-n]
-   (fn [^Session ws-session]
-     (let [send-ch (async/chan send-buffer-or-n)
-           remote  ^RemoteEndpoint (.getRemote ws-session)]
-       ;; Let's process sends...
-       (go-loop []
-         (if-let [out-msg (and (.isOpen ws-session)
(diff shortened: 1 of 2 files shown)


jetty/src/io/pedestal/http/jetty/websockets.clj
```diff
@@ -48,47 +64,22 @@
...
-           (do (try (ws-send out-msg remote)
-                    (catch Exception ex
-                      (log/error :msg "Failed on ws-send"
-                                 :exception ex)))
-               (recur))
+           (let [ws-send-ch (ws-send-async out-msg remote)
+                 result (async/<! ws-send-ch)]
+             (when-not (= :success result)
+               (log/error :msg "Failed on ws-send-async"
+                          :exception ex))
```
ddeaguiar, line 71: Should `ex` be `result`?

ohpauleez:
What's the rationale for removing the protocol extensions (a breaking change) and deprecating the protocol?
Users have an option of blocking synchronous calls or non-blocking asynchronous calls, and can use those building blocks to construct any system they wish.  Is is possible to leave the protocols as they are today while still correcting the blocking I/O from within go-loops?

Imagine a user who is using Pedestal's parts, but building up their own ChainProvider or Jetty integration and wants tight control over integration.  They should have access to both sync and async calls.

As another example, some integrations require tight control over thread allocation and only have a handful of static clients (eg: an internal web-service/microservice that is 'permanently' connected to a very small number of other internal services).  In those scenarios, the user may want to prioritize latency over throughput, preferring to use synchronous I/O, dedicated communication threads, and avoiding unbounded I/O threadpools entirely.  Or they might consider building an async service using core.async, but specifically want to size the worker pool based on the number of available cores and will avoid any use of unbounded threadpools, and is ok blocking workers.

ddeaguiar:
> I propose that the sync protocol was an accident, and that `WebSocketSend` and `WebSocketSendAsync` are implementation protocols that afford users of the client interfaces of `start-ws-connection` and `start-ws-connection-async` to send various types of value over the websocket core async channels. Neither of those should ever have used synchronous send (because Pedestal uses these protocols in go loops), and so the sync protocol should not have been created.

@aredington, I have no idea how common the usage of the sync protocol is but Paul is right that it's a breaking change. I can't speak to whether the sync protocol was an accident or not but I suppose that's irrelevant at this point because it's out there. For backwards compatibility sake what about leaving the sync protocol impl with a warning about usage?

ddeaguiar (approved):
Thanks @aredington!

PR #712, opened 2022-03-11
added missing expand-route to run-dev function to fix NPE

kreutter:
For some pedestal samples the `run-dev` method in `server.clj` did not call `io.pedestal.http.route/expand-routes` which result in a `NullPointerException` on each request. This problem occurs only in _dev mode_ when starting the example via `(run-dev)`.


bonkydog:
Hello, kreutter!

Please find instructions on how (and why) to do this here:

Thanks!
Brian

bonkydog:
Taken care of in https://github.com/pedestal/pedestal/pull/531

PR #713, opened 2022-05-09
allow `log/with-context` to accept non-map context arguments

philomates:
It seems that `io.pedestal.log/with-context` only works properly when provided a map value and not a reference to a map value.

For example
```clojure
(log/with-context attrs
  (log/info :request attrs))
```
gives you
```
Syntax error macroexpanding log/with-context at (io/pedestal/log_test.clj:10:7).
Don't know how to create ISeq from: clojure.lang.Symbol
```

By chance other arbitrary expressions "work" (because at macro-expansion time they are list data). So in the codebases I'm working with I often see:
```clojure
(log/with-context (identity attrs) ;; FIXME with-context macro doesn't like a symbol
  (log/info :request attrs))
```

On the one hand, only operating over concrete map values is inline with the `error`, `warn`, `info`, and other logging macros in `io.pedestal.log`. Yet at the same time the logic in `with-context` doesn't seem to explicit need a map value to work. Hence I'd like to propose a small change to allow users to do `(log/with-context some-var ...)` instead of `(log/with-context (identity some-var) ...)`

### approach i.

[the first commit](https://github.com/pedestal/pedestal/commit/806bca8f4f17bcd16cf236c5fb310487167301cb) proposes a change that just doesn't break but doesn't necessarily allow for macro-expansion time lookups of things like `::formatter` or `::mdc`.

### approach ii.

in the [second commit](https://github.com/pedestal/pedestal/commit/2dec8dde956ecf13e15b21ed3de6deca2643cc2f) I decided that it might be more comprehensive to shift more work from expansion time to evaluation time. I'm unsure if there is a particular reason for not doing this beforehand. I suspect there is, so I'm looking forward to the maintainer's thoughts


log/src/io/pedestal/log.clj
```diff
@@ -374,14 +374,14 @@
...
-  (let [formatter (::formatter ctx-map pr-str)]
-    (if (empty? ctx-map) ;; Optimize for the code-gen/dynamic case where the map may be empty
-      `(do
-         ~@body)
-      `(let [old-ctx# *mdc-context*
-             mdc# (or ~(::mdc ctx-map) (MDC/getMDCAdapter))]
-         (binding [*mdc-context* (merge *mdc-context* ~ctx-map)]
-           (-put-mdc mdc# mdc-context-key (~formatter (dissoc *mdc-context*
+  `(let [ctx-map-val# ~ctx-map]
+     (if (empty? ctx-map-val#) ;; Optimize for the code-gen/dynamic case where the map may be empty
```
philomates, line 378: I wonder how often this `(empty? ctx-map)` predicate is satisfied, especially in the previous impl when you had to explicitly pass `{}` for it to trigger. We could preserve this optimization at expansion time but it would complicate the implementation a bit with conditionals

ohpauleez: `(empty? nil)` returns true as well.  This portion of the logging library expects that most users will build a higher-level context construct on top of these calls (their own macro for sourcing and setting the MDC).
I think checking for `map?` is probably the best approach here, reducing the branching within the macro's generated code.

As a general rule of thumb, the logging utilities attempt to remove as much overhead from the logging interactions as possible.  It's common in high-throughput scenarios to see GC throughput limit application goodput -- one source of this can be excessive logging, forcing developers to decide between diagnostics and operational requirements.

Pedestal2 also introduces `with-mdc-copy` which avoids the `formatter` pieces and works against the MDC directly, correctly copying the MDC across thread boundaries.  In the systems I've been working with, this approach is more common because it's friendlier for structured log processors (like DataDog, etc.)

philomates: Cool, thanks for the context, it really helps me wrap my head around whats going/the constraints in flight.

> I think checking for map? is probably the best approach here, reducing the branching within the macro's generated code.

so essentially just go with https://github.com/pedestal/pedestal/pull/713/commits/806bca8f4f17bcd16cf236c5fb310487167301cb for a fix? And you think it is fine to then fail and fall back with the `::formatter` and `::mdc` lookups?

ohpauleez: Sorry for the confusion -- you want `nil` to short-circuit (as it already does):
If a developer's higher-level macro produced a `nil` context, you want this to fall through the correct code-gen path.

bonkydog:
Hello, Phillip!

Please find instructions on how (and why) to do this here:

Thanks!
Brian

log/src/io/pedestal/log.clj
philomates: revisiting this now. How does https://github.com/pedestal/pedestal/pull/713/commits/6a028e78d5cc4bd36fb6e5eeb31e3aae3f6f3ce2 look?

hlship:
Please reopen once you sign the CLA.

log/src/io/pedestal/log.clj
```diff
@@ -388,23 +388,25 @@
...
-                                                              :io.pedestal.log/formatter
-                                                              :io.pedestal.log/mdc)))
-           (try
-             ~@body
-             (finally
-               (-put-mdc mdc# mdc-context-key ((:io.pedestal.log/formatter old-ctx# pr-str)
-                                               (dissoc old-ctx#
-                                                       :io.pedestal.log/formatter
-                                                       :io.pedestal.log/mdc))))))))))
+  (if (or (symbol? ctx-map)
```
hlship, line 391: Limiting to just symbol here feels too constraining, as I could easily imagine wanting to invoke a function to generate the context map, rather than it being inline only.  As currently coded, if an expression is provided, it will be treated as an empty map, which will certainly confuse users.

hlship (changes requested):

log/src/io/pedestal/log.clj
philomates: Do you mean an expression like
```clojure
(log/with-context ((constantly {:a :b})) ["some" 'body])
```
or a var pointing to a function?:
```clojure
(let [gen-ctx-map-fn (constantly {:a :b})]
  (log/with-context gen-ctx-map-fn ["some" 'body]))
```

For the second it isn't possible (at macro expansion time) to know what is a function that the macro should expand into code that invokes it, and what is pointing to a base map. Doing it at eval time would work but go against the [suggestion to limit overhead](https://github.com/pedestal/pedestal/pull/713#discussion_r867978508).

For the first example, it would be possible to detect the expression at expansion time and have the macro emit code that shifts the work, that is, the evaluation of `(::formatter ctx-map pr-str)` and `(::mdc ctx-map)` expressions to normal runtime

hlship:
```
(log/with-context (create-context-from-request request) ....
```

The `(create-context-from-request ...)` form would fail the symbol? check.

philomates: my comment was a bit confusing, so now I'm even more confused. You thinking something like this https://github.com/pedestal/pedestal/pull/713/commits/180332087742b91f4abbffb025a7c890c6868d77 ?

vedang:
+1 for this to be merged into the main repo early. 

My use-case is as follows - I want the default formatter to print JSON, instead of `pr-str`, because it let's me push all logs to a structured logging endpoint. I do not want everyone using the code-base to have to specify `:io.pedestal.log/formatter json/write-value-as-string` in every log line they write. Moreover, I cannot guard against someone forgetting, and it will break my logging pipeline. 

The fix I have used is to implement a wrapper namespace over `io.pedestal.log`, with macros which just add the formatter to the map. This works fine for the logging macros, for example: 

```
(defmacro error [& keyvals]
  `(log/error ::log/formatter json/write-value-as-string ~@keyvals))
```

but it fails for the `with-context` macro because of the `Don't know how to create ISeq from: clojure.lang.Symbol` error. 

For example, this fails:
```
(defmacro with-context
  [ctx-map & body]
  (if (empty? ctx-map) ;; Optimize for the code-gen/dynamic case where the map may be empty
    `(do ~@body)
    `(log/with-context
       (assoc ~ctx-map ::log/formatter json/write-value-as-string)
       ~@body)))
```

Once the change described in this PR is merged in, I can rewrite my macro as:
```
(defmacro with-context
  [ctx-map & body]
  (if (empty? ctx-map) ;; Optimize for the code-gen/dynamic case where the map may be empty
    `(do ~@body)
    `(let [ctx-map# (assoc ~ctx-map ::log/formatter json/write-value-as-string)]
      (log/with-context ctx-map# ~@body))))
```

which does the job. 

Can you please merge this in soon? Or let me know if there is a better way to achieve what I want?

PR #717, opened 2022-06-17
Bump tajvm, treader and core.async to fix clj11 warnings

victoraldecoa:
Bump tajvm to 1.2.2, treader to 1.3.6 and core.async to 1.5.648 to fix clj11 warnings


mtsbarbosa:
It looks like we don't have much activity here this year
a pity, cause this PR would be pretty cool to come up with

Maybe we should consider migrating to a new forked repo

ddeaguiar (approved):

hlship:
Thanks for the changes, I've adapted them to the deps.edn build as #729 .

PR #721, opened 2022-08-09
Port to jetty 10.x and update deps to latest

ghaskins:
We intentionally hold back on tomcat 10.x as there is more porting work than to simply
bump the dependencies up.

It is suggested that if this is merged, the minor version should be bumped before the next release.

Signed-off-by: Greg Haskins <greg@manetu.com>

### DIFF
--- a/service/project.clj
+++ b/service/project.clj
@@ -16,31 +16,34 @@
   :scm "https://github.com/pedestal/pedestal"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies [[org.clojure/clojure "1.10.1"]
+  :dependencies [[org.clojure/clojure "1.11.1"]
 
                  [io.pedestal/pedestal.log "0.5.11-SNAPSHOT"]
                  [io.pedestal/pedestal.interceptor "0.5.11-SNAPSHOT"]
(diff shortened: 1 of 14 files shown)


hlship:
    
lease find instructions on how (and why) to do this here:
    
Thanks!

ghaskins:
CLA signed

service/project.clj
```diff
@@ -16,31 +16,34 @@
...
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies [[org.clojure/clojure "1.10.1"]
+  :dependencies [[org.clojure/clojure "1.11.1"]
 
                  [io.pedestal/pedestal.log "0.5.11-SNAPSHOT"]
                  [io.pedestal/pedestal.interceptor "0.5.11-SNAPSHOT"]
                  [io.pedestal/pedestal.route "0.5.11-SNAPSHOT"]
 
+                 [org.ow2.asm/asm "9.3"]
```
souenzzo, line 25: Why does it is required?

ghaskins: I actually no longer recall.  It was probably either one of two things, both related to transitive deps:

1) needed to be updated for the PR to work
2) there was a CVE 

IIRC, it was the former (1).  I can try to look later.

hlship:
Replaced by #762

PR #723, opened 2022-09-09
Convert from lein to deps.edn build

hlship:
This PR has gotten a bit bigger than I'd originally envisioned:
- Updated all the copyrights
- Replace Leiningen build with Clojure CLI build
- Replace release/deploy scripts with Clojure build tools
- Removed a dependency cycle by boosting the tests to a new sub-project
- New build tool to handle updating the version number (across all relevant deps.edn and project.clj files)
- New build tool to handle the logic for advancing a version number


aws/deps.edn
```diff
@@ -0,0 +1,38 @@
...
+; By using this software in any fashion, you are agreeing to be bound by
+; the terms of this license.
+;
+; You must not remove this notice, or any other, from this software.
+
+{:paths ["src"]
+ :deps {org.clojure/clojure {:mvn/version "1.10.1"}
+        io.pedestal/pedestal.log {:mvn/version "0.5.11-SNAPSHOT"}
+        io.pedestal/pedestal.interceptor {:mvn/version "0.5.11-SNAPSHOT"}
+
```
hlship, line 16: There's a new build tool to keep these version numbers synchronized.

aws/deps.edn
```diff
@@ -0,0 +1,38 @@
...
+        ;; Bring AWS libs inline with Pedestal Service
+        com.fasterxml.jackson.core/jackson-core {:mvn/version "2.9.9"}
+        com.fasterxml.jackson.dataformat/jackson-dataformat-cbor {:mvn/version "2.9.9"}
+        commons-codec/commons-codec {:mvn/version "1.12"}
+        joda-time/joda-time {:mvn/version "2.10.2"}}
+
+ :aliases
+ {:local
+  {:override-deps {io.pedestal/pedestal.log {:local/root "../log"}
+                   io.pedestal/pedestal.interceptor {:local/root "../interceptor"}}}}}
```
hlship, line 38: But this trick means you don't need to care about dependency versions, as it can just get them from local workspace, they way you'd like.

tests/deps.edn
```diff
@@ -0,0 +1,78 @@
+; Copyright 2022 Cognitect, Inc.
```
hlship, line 1: So a big part of this change is to create a tests project that contains all the tests; previously there was some very dodgy dependencies that were almost circular. Unfortunately, this also meant moving all the test namespaces into the new project as well.

.github/workflows/ci.yml
```diff
@@ -15,12 +15,16 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@v1
         with:
-          java-version: '1.8'
+          java-version: '1.9'
```
hlship, line 18: Something referenced java.lang.ProcessBuilder, while is first available in 1.9.

bonkydog (commented):
nice! :rocket:

bonkydog (approved):
:rocket:

PR #746, opened 2023-05-10
Fix Pedestal Issue #610

bonkydog:
https://github.com/pedestal/pedestal/issues/610

`url-for` throws and exception when a context path is supplied and used for a route with a path parameter in the first segment.

The logic appeared to assume that path segments are always strings and was using `empty?` to check for the empty string.  `empty?` expects a sequence.  A keyword is not a sequence, so empty was throwing an exception.

### DIFF
--- a/route/src/io/pedestal/http/route.clj
+++ b/route/src/io/pedestal/http/route.clj
@@ -247,7 +247,6 @@
         path-parts (do (log/debug :in :link-str
                                   :path-parts path-parts
                                   :context-path-parts context-path-parts)
-                       ;;(concat context-path-parts path-parts)
                        (cond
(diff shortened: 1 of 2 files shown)


route/src/io/pedestal/http/route.clj
```diff
@@ -249,7 +249,7 @@
                                   :context-path-parts context-path-parts)
                        ;;(concat context-path-parts path-parts)
```
hlship, line 250: This comment should be deleted.

hlship (approved):
Maybe delete that comment before merging.

route/src/io/pedestal/http/route.clj
bonkydog: ah, thanks!

hlship:
Above changes accidental, this was fixed in 0.6.0.

PR #748, opened 2023-06-01
Update dependencies to cover a number of CVEs

hlship:


bonkydog (approved):
Ahhh, so good to have this CVE check added to CI!

PR #749, opened 2023-06-01
Add documentation and further tests related to routing

hlship:
Also: 
- Fixed SLF4J warnings when generating codox or running tests
- Remove depedency on org.clojure/core.incubator
- Fix a bug in codox task


bonkydog:
👀

bonkydog (approved):
Thanks for the docstrings and better names. Much more comprehensible!

hlship:
Yes, I got sick of having to figure it out each time because the names were so confusing.  It's all still more complicated than it needs to be.

PR #762, opened 2023-10-06
Update to Jetty 11 and use of the jakarta.* packages

hlship:
This also removes support for tomcat; I put in a tag (`hls/prior-to-scrap-tomcat`) so that we can recover the deleted code, should we decide to resurrect it.


jetty/src/io/pedestal/http/jetty.clj
```diff
@@ -177,23 +148,24 @@
...
         ssl (when (or ssl? ssl-port h2?)
               (ssl-conn-factory (assoc options :alpn alpn)))
         http-connector (when port
                          (doto (ServerConnector. server (into-array ConnectionFactory
+
                                                                     (remove nil? [http http2c])))
                            (.setReuseAddress reuse-addr?)
                            (.setPort port)
                            (.setHost host)))
+        ;; TODO: This is looking to be _very_ different in Jetty 11!
```
bonkydog, line 168: probably delete this now

bonkydog (approved):
Looks mostly great, but I left a few notes.

brown131:
When will this be included in a new release?

hlship:
Yes, it will likely be the focus of the 0.7 release, along with the new documentation.

PR #766, opened 2023-10-23
Move documentation from pedestal-docs to pedestal

hlship:
As part of the move to Antora as the documentation build tool, this PR moves documentation from the pedestal-docs repo to the main pedestal repo.

This will allow documentation changes to be synchronized with code changes.

Along the way, examples were updated to Pedestal 0.6.1, and much content was improved and reorganized.

PR #769, opened 2023-10-27
Performance improvements and address latest CVEs

hlship:
io.pedestal.interceptor.chain: added macro `bind`, and function `on-enter-async`.

A few scattered other improvements, such as changing usages of `update-in` to `update`.

Documentation will come once #766  has merged.


hlship:
Build may fail due to CVEs, which can be addressed later.

hlship:
Getting some spurious test failures in CI, related to bound variables. This is new territory, as there wasn't much testing of bound vars prior. Still, the failure is very odd (a bound var's bound value seems to persist after it is no longer bound).

ohpauleez:
Thanks for taking a look at the performance of the core interceptor chain!

It would be helpful to understand the motivation of some of these changes.
For example, Pedestal's current public contract is that all interceptors that are executed are added to the stack.  Sophisticated interceptors (and debugging interceptors) can use this information to understand how a "chain" ended up in a certain state.  The proposed changes here presents a breaking change to that contract.

The changes here also open up the public contract of the chain (in a way that I think is potentially dangerous or limiting for future evolution).  Historically "binding" has remained an internal detail because it was to ensure binding conveyance in the face of async execution (not as a general public mechanism for establishing "bindings" in the context).  Users already have a way to establish bindings in the context by `assoc`ing data into the context, and a way to pass bindings forward across async boundaries with `binding`. 

I would strongly suggest the benchmarks be measured in JMH instead of Criterium.  Looking at latency alone without seeing throughput (or allocation overhead, or GC churn) can often be quite misleading in these kinds of systems.  These changes also introduce a new set of conditional checks within the hot loop, so you're at the whim of branch prediction -- running JMH with `perf` will quantify the impact of branch predicition misses.  Generally speaking, a hot loop should be free of conditionals where possible (or include a single conditional that checks/confirms the sentinel check).

- - - - -

As a general thought, I would leave the Pedestal1 chain mostly as-is and make small tweaks where possible but not change any aspect of the public contract.  There are _much_ better approaches to designing this interceptor chain execution, but they all involve breaking changes -- so if you're going to make a breaking change, you might as well make the one that nets the best return :)  (ie: Pedestal2 can process millions of requests per second on a single box).  All that said, if you're focused to hit a certain latency and/or throughput target on Pedestal1, it would be good to know what that is and I can do my best to pass along things that have been tried and approaches that have worked in practice.

interceptor/src/io/pedestal/interceptor/chain.clj
```diff
@@ -80,205 +85,215 @@
...
+  active-interceptor: interceptor invoked
+  continuation: the continuation map
+  "
+  [context-ch execution-context active-interceptor continuation]
+  (log/debug :in 'handle-async
+             :execution-id (::execution-id execution-context)
+             :interceptor (name active-interceptor))
+  (let [context' (prepare-for-async execution-context)]
+    (async/take! context-ch
+                 (fn [new-context]
```
hlship, line 135: This path can lead to :entrer-async callbacks being invoked again.  Needs to dissoc :enter-async.

hlship:
I haven't looked at Pedestal 2.

The documentation about :bindings (http://pedestal.io/reference/context-map#_bindings) clearly indicates its there for interceptors to use.  I actually haven't changed that, just provided convenience macros to make it easier to use.  That documentation dates back to 2016 (https://github.com/pedestal/pedestal-docs/commit/6359650d0e3c54901179bb74ab7036ba1978bf48).

In terms of queue vs. stack; the ::stack part was undocumented behavior and the documentation about ::queue is purposely vague.

It sounds like you have a definition of a "contract" for this behavior that isn't captured in the code, documentation, or tests.

My general approach to performance: pre-compute what can be pre-computed, minimize work, and (as in this case) minimize changes to persistent collections.  Certainly, the code would be less branchy if we could discard `execute-only` (that's the majority of the branches you've discussed).

hlship:
```
+-----------------------------------+----------------+
| dependency                        | status         |
+-----------------------------------+----------------+
| http2-common-11.0.16.jar          | CVE-2023-44487 |
| http2-server-11.0.16.jar          | CVE-2023-44487 |
| jetty-io-11.0.16.jar              | CVE-2023-44487 |
| jetty-server-11.0.16.jar          | CVE-2023-44487 |
| websocket-core-client-11.0.16.jar | CVE-2023-44487 |
| websocket-core-server-11.0.16.jar | CVE-2023-44487 |
+-----------------------------------+----------------+
```

We'll address these in a different PR.

hlship:
Either my early benchmarks were not valid, or the other modest changes scattered about were more important than the larger changes to interceptor.chain; still getting better numbers after reverting most of the interceptor.chain changes.

bonkydog (approved):
Great! 🚀

Issue #780, opened 2023-12-10
Add configuration for linting using clj-kondo


vedang:
Hello maintainers,

It would be great if Pedestal provided `clj-kondo` configuration[^1] in the `io.pedestal` jars which introduce macros. This makes using the macro easier and ensures that `clj-kondo` does not throw false warnings. 

[^1]: Providing `clj-kondo` configuration as part of the project: https://github.com/clj-kondo/clj-kondo/blob/be6efa075a0a2b903ab220ce7c7e61a242bd9126/doc/config.md#exporting-and-importing-configuration

hlship:
Thanks for putting this together; I've taken it further by properly bundling the config files and hooks, integrating a GitHub action to run the linter, and fixing dozens of linter warnings and errors; this was in PR #816

PR #781, opened 2023-12-10
Add clj-kondo linting configuration for `interceptor.error/error-dispatch` and `log/<level>` macros

vedang:
This commit adds a `clj-kondo.exports` folder as explained in `clj-kondo` documentation. Ideally, if this is bundled into the jar when library versions are cut, then users will be able to import the config in their own projects[^1]. This commit shows the example layout of the config, but more work needs to be done to bundle this into `resources` available in the final jars.

We use the `:analyze-call` hook[^2] to rewrite the call to `error-dispatch` as an anonymous function with a `case` statement. This matches the spirit of the macro-expansion and works for linting purposes.

Related to: #780

[^1]: Exporting and importing clj-kondo configuration: https://github.com/clj-kondo/clj-kondo/blob/be6efa075a0a2b903ab220ce7c7e61a242bd9126/doc/config.md#exporting-and-importing-configuration 
[^2]: An explanation of the `:analyze-call` hook:
https://github.com/clj-kondo/clj-kondo/blob/be6efa075a0a2b903ab220ce7c7e61a242bd9126/doc/hooks.md#analyze-call

### DIFF
--- a/clj-kondo-resources/clj-kondo.exports/io.pedestal/pedestal.log/config.edn
+++ b/clj-kondo-resources/clj-kondo.exports/io.pedestal/pedestal.log/config.edn
@@ -1,4 +1,4 @@
-; Copyright 2022 Cognitect, Inc.
+; Copyright 2024 Nubank NA
 
 ; The use and distribution terms for this software are covered by the
 ; Eclipse Public License 1.0 (http://opensource.org/licenses/eclipse-1.0)
--- a/clj-kondo-resources/clj-kondo.exports/io.pedestal/pedestal.log/config.edn
+++ b/clj-kondo-resources/clj-kondo.exports/io.pedestal/pedestal.log/config.edn
@@ -9,5 +9,10 @@
 ;
 ; You must not remove this notice, or any other, from this software.
 
-{:config-paths ["clj-kondo-resources/clj-kondo.exports/io.pedestal/pedestal.interceptor"
-                "clj-kondo-resources/clj-kondo.exports/io.pedestal/pedestal.log"]}
+{:hooks
+ {:analyze-call
+  {io.pedestal.log/trace hooks.io.pedestal.log/log-expr
+   io.pedestal.log/debug hooks.io.pedestal.log/log-expr
+   io.pedestal.log/info  hooks.io.pedestal.log/log-expr
+   io.pedestal.log/warn  hooks.io.pedestal.log/log-expr
+   io.pedestal.log/error hooks.io.pedestal.log/log-expr}}}
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -1,4 +1,4 @@
-; Copyright 2022 Cognitect, Inc.
+; Copyright 2024 Nubank NA
 
 ; The use and distribution terms for this software are covered by the
 ; Eclipse Public License 1.0 (http://opensource.org/licenses/eclipse-1.0)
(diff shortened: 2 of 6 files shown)


hlship:
Thanks for the PR; I have to check if you've signed the CLA, and we need to add copyright headers to source files.  I don't see a reason why we shouldn't package the clj-kondo support.

hlship:
I see you've signed the CLA.

vedang:
I'll add the copyright headers to the source files and update this PR. Please note that you will still need to adjust the packaging scripts to ensure that the `clj-kondo.exports` folder is on the classpath, for the config to be useful to users for `pedestal` libraries.

vedang:
I've also added some clj-kondo configuration for the log/trace, log/debug ... macros. I found these really useful after making a mistake and wrote an odd number of values for one of the macros.

clj-kondo-resources/clj-kondo.exports/io.pedestal/pedestal.log/config.edn
```diff
@@ -0,0 +1,18 @@
+; Copyright 2022 Cognitect, Inc.
```
stig, line 1: Should this be 2023?

vedang: I copied it from one of the existing files and didn't catch this :( . Will wait for @hlship to tell me what the right value should be, or if I can just leave it as-is, before making a change.

hlship: (c) 2024 Nubank NA

hlship:
I'll take care of build changes after we merge this PR.

vedang:
I've pushed the new changes with the updated copyright value. I have also rebased the changes on top of the latest master.

hlship (approved):

adityaathalye:
Oh neat! I was just about to write my own clj-kondo config for `error-dispatch`, and paused to wonder why the library does not already have support. Thanks @vedang and @hlship for this!

That said, I just switched to the 0.7.0-SNAPSHOT version to try and get this improvement. I'm not sure why it's not working.

- My deps.edn has `io.pedestal/pedestal.jetty {:mvn/version "0.7.0-SNAPSHOT"}`
- I followed clj-kondo's Import instructions (I have clj-kondo v2024.02.12).
- My project has prismatic/schema and copy-configs reports action taken only for that library.
  ```shell
  $ clj-kondo --lint "$(clojure -Spath)" --copy-configs --skip-lint
  Configs copied:
  - .clj-kondo/prismatic/schema
  linting took 259ms, errors: 0, warnings: 0
  ```
- A full lint appears to skip over pedestal entirely.
  ```
  $ clj-kondo --lint "$(clojure -Spath)" --dependencies --parallel --copy-configs
       | grep pedestal   # finds nothing
  ```
- Linting my source directory produces these errors, for usage of the `error-dispatch` macro.
  ```
  $ clj-kondo --lint src
  the-path/to/pedestal_component.clj:99:26: error: Unresolved symbol: ctx
  the-path/to/pedestal_component.clj:99:30: error: Unresolved symbol: ex
  linting took 150ms, errors: 2, warnings: 0
  ```
- Deleting the project's .clj-kondo/.cache and re-running the steps does not alter the situation.

I must be missing something blindingly obvious, but I can't figure out what.

hlship:
No the issue here is that this PR handles linting when working on Pedestal, but I neglected add the final stage that packages the linting config with the libraries.

adityaathalye:
Thanks for looking into this! It was late night for me and I stopped just shy of looking inside the JAR. I figured I would investigate for a bit again in the morning with fresh eyes.

PR #783, opened 2023-12-15
Add new pedestal-metrics library, with functional interface built on top of Micrometer

hlship:
This leaves the existing metrics support into io.pedestal.log unchanged, for the meantime.

This is a first step; further work: 
- documentation & examples 
- a standard timing interceptor
- adapt new code to DropWizard (help with transition from Pedestal 0.6)


bonkydog (approved):
🚀

hlship:
I renamed this PR to make more sense in light of what eventually happened, so it reads better in the release notes; micrometer ended up being an intermediate step.

PR #784, opened 2023-12-16
Add implemention of MetricSource SPI for DropWizard/Codahale MetricRegistry

hlship:
This is a follow-on to #783 that provides a way to use the new io.pedestal.metrics namespace with DropWizard/Codahale metrics, as a transition path for people upgrading to Pedestal 0.7 with an investment in the old metrics.


bonkydog (approved):
:rocket:

PR #785, opened 2023-12-20
Allow global override of event map formatter

### LINKED ISSUE #666: Allow to customize the "default" logger
# Description

`io.pedestal.log` currently support to use custom formatted by parameter.

By default, it will fall into `pr-str` and it's not customizable.

- https://github.com/pedestal/pedestal/blob/master/log/src/io/pedestal/log.clj#L271
- https://github.com/pedestal/pedestal/blob/master/log/src/io/pedestal/log.clj#L289
- https://github.com/pedestal/pedestal/blob/master/log/src/io/pedestal/log.clj#L391

## Expected Behavior

Should be possible to configure the logger formatter from java properties

like `-Dio.pedestal.log.default-formatter=my.custom.log/encoder`

If I'm running in AWS, i can format the output as JSON because CloudWatch can understand it

If I'm running on my machine, i can log as EDN because it's easier to read

## Actual Behavior

Logs are always printed with EDN, even when using an JSON logger appender.
```javascript
{"level": "INFO", "message": "{:msg \"GET /hello\", :line 20}" ...}
```

I need to create my custom log namespace per-project and handle that

## Pedestal version

0.6.0

hlship:
Fixes #666

The use of a default formatter applies to MDC events as well.

This also deprecates io.pedestal.log/with-context-kv, which is untested and looks broken.

### DIFF
--- a/log/src/io/pedestal/internal.clj
+++ b/log/src/io/pedestal/internal.clj
@@ -9,8 +9,9 @@
 ;
 ; You must not remove this notice, or any other, from this software.
 
-(ns  ^:no-doc io.pedestal.internal
-  "Internal utilities, not for reuse, subject to change at any time.")
+(ns ^:no-doc io.pedestal.internal
+  "Internal utilities, not for reuse, subject to change at any time."
+  (:require [clojure.string :as str]))
 
 (defn vec-conj
   [v value]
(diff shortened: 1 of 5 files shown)


log/src/io/pedestal/internal.clj
```diff
@@ -19,3 +20,33 @@
...
+                           (println-err (format "WARNING: Symbol %s (from %s %s) does not exist"
+                                                s
+                                                from-source from-name))))
+                       (catch Exception e
+                         (println-err (format "ERROR: Could not resolve symbol %s (from %s %s): %s"
+                                              s
+                                              from-source from-name
+                                              (ex-message e)))))))]
+    (or (resolver "JVM property" property-name (System/getProperty property-name))
+        (resolver "environment variable" env-var (System/getenv property-name)))))
```
vedang, line 52: `(resolver "environment variable" env-var (System/getenv property-name))` should be 
`(resolver "environment variable" env-var (System/getenv env-var))` right?

hlship: Good catch!  Manual testing only goes so far, and environment variables are immutable globals.

bonkydog (approved):
:rocket:

PR #788, opened 2023-12-21
Fix minor initialization bug in ClojureVarServlet

### LINKED ISSUE #755: There is no test for the ClojureVarServlet
Although Pedestal supplies the ClojureVarServlet (used when deploying a Pedestal application as a WAR file) it does not actually test it in any way.

*Update*: Found the error (IndexOutOfBounds) while testing, and corrected some invalid documentation.

hlship:
ClojureVarServlet is used when deploying a Pedestal service inside a container (rather than the normal approach, to use a container embedded in the application).

Updated some incorrect documentation and found a bug where if there was no slash in the function name (which is supposed to be fully qualified) it would throw an IndexOutOfBoundsException.

Fixes #755

### DIFF
--- a/service/deps.edn
+++ b/service/deps.edn
@@ -9,7 +9,9 @@
 ;
 ; You must not remove this notice, or any other, from this software.
 
-{:paths ["src" "target/classes"]
+{:paths ["src"
+         "java"
+         "target/classes"]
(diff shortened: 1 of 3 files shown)


everaldofilho (approved):
LGTM

PR #796, opened 2024-01-05
Add dev-mode printing of routing table

hlship:
When in dev-mode, the `routes-from` macro will now print a sorted, formatted table from the expanded routing specification:

```
Routing table:
┏━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Method ┃ Path   ┃ Name                    ┃
┣━━━━━━━━╋━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ :get   ┃ /hello ┃ :io.example.hello/hello ┃
┗━━━━━━━━┻━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

This is output at startup, and on any subsequent request where the routing table has visibly changed (any method, path, or route name has changed).

### DIFF
--- a/tests/test/io/pedestal/http/route_repl_test.clj
+++ b/tests/test/io/pedestal/http/route_repl_test.clj
@@ -45,21 +45,50 @@
         alt-routes #{["/bye" :get #'bye-handler :route-name ::bye]}]
     (is (fn? f))
 
-    (is (= (simplify (route/expand-routes sample-routes))
-           (simplify (f))))
+    (let [out-str (with-out-str
+                    (is (= (simplify (route/expand-routes sample-routes))
+                           (simplify (f)))))]
+      (is (= "Routing table:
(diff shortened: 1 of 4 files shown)


tests/test/io/pedestal/http/route_repl_test.clj
```diff
@@ -45,23 +45,64 @@
         alt-routes #{["/bye" :get #'bye-handler :route-name ::bye]}]
     (is (fn? f))
 
-    (is (= (simplify (route/expand-routes sample-routes))
-           (simplify (f))))
+    (let [out-str (with-out-str
```
hlship, line 48: I decided to link the printing of the routing table directly to generated _function_ so that it only occurs when uring `route-from`. There _might_ be cases of production applications that still use a function to provide the specification.

PR #803, opened 2024-02-05
Fix: Pedestal.log depends on Pedestal.common

vedang:
Add a dependency so that `io.pedestal.internal` resolves correctly


hlship (approved):

hlship:
Thanks for the fix!  Could you look into GPG signing your commits, for any future contributions?

vedang:
Sure 👌 I'll do that for all future contributions

PR #866, opened 2024-08-19
Support Jetty 12 ee10

terjesb:
Support Jetty 12 via pedestal.jetty-12-ee10 module, along existing Jetty 11 pedestal.jetty module.

For now, both modules uses the same io.pedestal.http.jetty namespace and the same :io.pedestal.http/type :jetty key. So one should only include one of the modules in a project.


hlship:
I'm checking this out now ...

There's some build conflicts that need fixing. I'll look into that.

hlship:
So, I'd prefer to see this as a different namespace, .jetty12 (and thus corresponding keyword, :jetty12).  I don't think that having two libraries that defining conflicting versions of the same namespace is maintainable.

Also, I'm sad that `clj -T:tests run` ends up dropping all the nice ANSI colors, but I can live with it.

terjesb:
The latest version uses the .jetty12 namespace and corresponding keyword, :jetty12.

hlship:
We've been having discussions about how to best serve the needs of Nubank and the wider Pedestal community; our growing consensus is a bit of a reversal --- back to your original approach, of upgrading the pedestal.jetty library from Jetty 11 to Jetty 12.

hlship:
That's made in the context of Jetty 11's EOL being in Jan 2025.

terjesb:
> back to your original approach, of upgrading the pedestal.jetty library from Jetty 11 to Jetty 12.

I've updated the branch accordingly. :jetty is now Jetty 12. First, I switched Jetty 11 to :jetty11 and changed :jetty12 to :jetty. Then I removed :jetty11 with Jetty 11 support altogether.

observer-test fails in the build ci for 1.11, but it seems to work locally on my machine even when I override-deps with 1.11.4. This test doesn't seem to use any Jetty support, so I don't understand this yet.

hlship:
Thanks for the effort; I definitely used some references from this PR in the PR I created.

PR #869, opened 2024-09-28
Rework routing protocols and implementations

hlship:
A major improvement here is that it is now possible to assemble the routing table from multiple sources.


route/src/io/pedestal/http/route/specs.clj
```diff
@@ -156,25 +157,55 @@
...
+;; -- ROUTING FRAGMENT --
+
+(s/def ::routing-fragment #(satisfies? types/RoutingFragment %))
+
+;; TODO: Can you spec protocol methods?
+
+(s/def ::fragment-routes (s/coll-of ::fragment-routing-entry))
+
+(s/def ::fragment-routing-entry (s/keys
+                         :req-un [::path
```
souenzzo, line 169: This indentation is too ugly!

Issue #872, opened 2024-10-10
developing-at-the-repl.adoc should be re-written around the embedded template


hlship:
It starts with "use `lein ...` to create your project" which is no longer viable.

hlship:
The document was retired in 0.8, in favor of https://pedestal.io/pedestal/0.8/guides/live-repl.html.

Fixed in c7a18ae4b19d73c7bc0c61928ec45f0cf9526e83

PR #873, opened 2024-10-10
Make Sawtooth the default router

hlship:


docs/modules/guides/pages/developing-at-the-repl.adoc
```diff
@@ -294,7 +294,7 @@ to `server.clj`:
 (defn recognize-route
   "Verifies the requested HTTP verb and path are recognized by the router."
   [verb path]
-  (route/try-routing-for (table-routes service/routes) :prefix-tree path verb))
+  (route/try-routing-for (table-routes service/routes) path verb))
```
hlship, line 297: This is not accurate, but #872 will have to address it.

PR #887, opened 2025-01-29
Extend routing to cover exposing file system and/or classpath resources

hlship:


docs/modules/reference/pages/resources.adoc
```diff
@@ -0,0 +1,137 @@
...
+route name collisions.
+
+The :prefix option should start with a "/" and not end with one.
+
+For the :index-files? option, Pedestal will identify requests that map to a directory, not to a specific file.
+Pedestal will search for one of the following:
+
+- `index.html`
+- `index.htm`
+- Any file whose name starts with `index.`
```
marcobiscaro2112, line 78: We could make this a numbered list if it tries to find these files in the order (which I assume it does; if that is not the case, we should clarify)

docs/modules/reference/pages/resources.adoc
```diff
@@ -0,0 +1,137 @@
...
+
+If any of these are found, then the request is treated as if it were to the matching file.
+
+The :resource-route option should not start or end with a slash; to restrict access to just classpath files within the `org.example.myapp.public` namespace, the option should be "org/example/myapp/public".
```
marcobiscaro2112, line 82: Should be `resource-root`?

docs/modules/reference/pages/resources.adoc
```diff
@@ -0,0 +1,137 @@
...
+[options=header,cols="1,1,^1,2p"]
+|===
+| Option | Default | Only | Description
+
+| :prefix          | "/public" |          | The path prefix for the exposed resources.
+| :root-path       | -         | file     | Path to root directory to expose.
+| :resource-root   | -         | resource | Prefix on resources, limiting access to a safe subset of the classpath
+| :classloader     | -         | resource | Classloader used to find resources, or nil for the default class loader
+| :index-files?    | true      | file     | If true, paths that map to directories may return the index file for that directory.
+| :cache?          | true      |          | If true, then resource data is cached to in memory to speed up requests.
```
marcobiscaro2112, line 53: This should probably default to false since the in-memory cache is unbounded (and hot resources are likely to be loaded from the OS page cache anyway)

service/src/io/pedestal/http/resources/impl.clj
```diff
@@ -0,0 +1,149 @@
...
+  (into-array OpenOption [StandardOpenOption/READ]))
+
+(defn- should-stream?
+  [request stream-size]
+  (let [{:keys [servlet-response]} request
+        buffer-size-bytes (if servlet-response
+                            (.getBufferSize ^HttpServletResponse servlet-response)
+                            ;; The comment: Assume 1500 MTU (https://en.wikipedia.org/wiki/Maximum_transmission_unit)
+                            ;; (originally copied from io.pedestal.http.ring-middlewares/fast-resource)
+                            ;; Q: Then why is the default 1460?
```
marcobiscaro2112, line 37: Probably to account for 20 bytes for TCP header + 20 bytes for IP header (assuming IPv4)

docs/modules/reference/pages/response-bodies.adoc
```diff
@@ -23,6 +24,14 @@ type of the value in the :body key of the xref:response-map.adoc[].
 | {core_async} channel | nil | see below | yes
 |===
 
+[#nio-channel]
+== nio Channel
+When the response body is a ByteBuffer or an NIO Channel, then Pedestal will convert
```
marcobiscaro2112, line 29: Why include ByteBuffer as well? We should probably only do it for Channel (there is no async operation to read from a ByteBuffer, as it is already mapped to memory)

marcobiscaro2112 (commented):
Some comments about the docs (I have not reviewed the impl yet)

docs/modules/reference/pages/resources.adoc
hlship: I'd prefer to leave this defaulted to true, as that's sensible for most applications that may have dozens, or even hundreds of resources to be exposed.  An application that have thousands of resources, or resources that might change during execution, should disable the cache, and perhaps, implement their own caching strategy as an interceptor around the routes.

docs/modules/reference/pages/response-bodies.adoc
hlship: I'm not sure, just maintaining parity with a lot of stuff that predates me.  
https://github.com/pedestal/pedestal/blob/a7ce6c0f5da45b7db45c9dc7aded9e80da808b4e/service/src/io/pedestal/http/impl/servlet_interceptor.clj#L148

https://github.com/pedestal/pedestal/blob/a7ce6c0f5da45b7db45c9dc7aded9e80da808b4e/service/src/io/pedestal/http/container.clj#L30

https://github.com/pedestal/pedestal/blob/a7ce6c0f5da45b7db45c9dc7aded9e80da808b4e/jetty/src/io/pedestal/http/jetty/container.clj#L48

hlship: So that looks like Jetty, at least, will support ByteBuffer to async operations.

PR #905, opened 2025-03-30
fix broken cond-> link in the guides

thiagooak:
I don't know how to generate the docs to check if this is correct. But it is rendering incorrectly in the website. I have a hunch that this will fix the problem (I'm assuming there is a regex somewhere that needs a trailing space to work)

https://pedestal.io/pedestal/0.7/guides/hello-world-content-types.html

image


hlship:
I'll take a peek at this soon; basically, things get *very very complicated* with multiple layers of interpretation and reinterpretation of non-alphabetic characters.

hlship:
Fixed by dde416b66a9580b45c0a72bc9c8e0dc8d1751e4e

PR #914, opened 2025-04-29
Unify Component and Pedestal Interceptors

hlship:
This makes it easy to define components, as Clojure records, that will transform into Interceptors (via the io.pedestal.interceptor/IntoInterceptor protocol) as needed (when added as a interceptor, or as an interceptor inside a route).

This opens up a viable alternative to the "big ball of of mud" approach to Component dependencies in a Pedestal application.


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 
 ## 0.8.0 - UNRELEASED
 
-The main focus of this release is to improve routing and upgrade to Jetty 12.
+The main focii of this release is to improve routing and upgrade to Jetty 12.
```
souenzzo, line 8: unexpected change//mistype

I think you use vim keybindings ;)

PR #918, opened 2025-05-16
Make level of cors interceptor logging configurable

### LINKED ISSUE #916: The allow-origin interceptor (CORS) produces too much log output
My logs are absolutely full of `cors request processing` and `cors response processing`, both INFO-level log messages from the `allow-origin` interceptor. It seems to me like they should have been DEBUG log calls instead of INFO as they are  incredibly spammy (they log twice for every single HTTP request) and not really needed 99.9% of the time. 

Merely changing these two log calls to DEBUG would remove the spam and would make me (and possibly others) much more happy with Pedestal.

hlship:
Fixes #916 

To fix this, I needed to retool some of the logging, to support determining the logging level at runtime.  Took some work to ensure this didn't change the normal case where the logging level is known at macro expansion time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ Other changes:
 - New functions and macros:
   - `io.pedestal.test/create-responder` - useful piece needed in most tests
   - `io.pedestal.interceptor/definterceptor` - easily create component records that transform into interceptors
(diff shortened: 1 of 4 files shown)


hlship:
Adding "breaking change" label ... but if the level of logging here breaks your system, I have questions. Anyway, you can configure it back to level :info if you need to.

PR #933, opened 2025-06-30
Include a prettier index.html page as part of the embedded template

das-monki:


hlship (commented):
This looks good, but I think there needs to be some additional changes to embedded to ensure the new files are packaged into the created project.

hlship:
Please retry the CLA signing procedure; it looks like it worked for some others recently.

PR #946, opened 2025-07-17
Interceptor error handling: don't include nested exceptions as :exception in ex-data

### LINKED ISSUE #640: ex-info's thrown by Pedestal contain problematic redundant data
# Description
ex-info's thrown by the interceptor chain contain data with two pieces of redundant information.  This redudancy causes huge output with the default exception printing machinery in clojure.core
This means that an ex-info with 2 chained causes (three exceptions total) will have 7 exceptions printed: 4 for the pedestal wrapper in front of the 3 exceptions, then the 3 exceptions printed _within the ex-data_.

The problematic info in ex-data:
1) `:exception`  -- this represents the cause, but it is already properly visible as the chained 
exception cause.  `pr` upon a exception from the interceptors will end up printing the cause twice, once as the outermost's cause, and once within the ex-data of the outermost exception.
2) the ex-data of the cause is re-merged in the top-level ex-data, which confuses the provenance (who threw that data?)

https://github.com/pedestal/pedestal/blob/621b6896aab6582499e0076bd9c73d5aeb01dae9/interceptor/src/io/pedestal/interceptor/chain.clj#L36-L41

## Expected Behavior
user exceptions have data, pedestal wraps an ex-info that chains the cause, with ex-data containing pedestal-specific augmentations

## Actual Behavior
`pr` of an interceptor-thrown exception

```
{:via
 [{:type clojure.lang.ExceptionInfo,
   :message
   "java.lang.NullPointerException in Interceptor :REDACTED - ",
   :data
   {:execution-id 4128,
    :stage :enter,
    :interceptor :REDACTED,
    :exception-type :java.lang.NullPointerException,
    :exception
    #error {:cause nil, :via [{:type java.lang.NullPointerException, :message nil, :at [java.util.UUID fromString "UUID.java" 192]}], :trace [[java.util.UUID fromString "UUID.java" 192] .....truncated

## Steps to reproduce
Throw an exception or ex-info in an interceptor :enter method, then `pr` the exception

# Environment

## Operating System (including version).
MacOS 10.14
## Your current Leiningen or Boot version (`lein --version` or `boot --version`)
2.9.1
## Pedestal version
0.5.7

hlship:
Fixes #640

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@
 - Fixed reloading behavior when namespaces are reloaded via [clj-reload](https://github.com/tonsky/clj-reload)
 - Server-Sent Events have been changed; fields are now terminated with a single `\n` rather than a `\r\n` (both
   are acceptible according to the SSE specification)
+- Exceptions in interceptors:
+  - The caught exception is now the `ex-cause` of the exception provided (in earlier releases, it was the :exception key of the data)
+  - The logic for when to suppress an exception thrown from the error handling interceptor has been simplified: always suppress except when the interceptor rethrows the exact error passed to it
 - `io.pedestal.test` has been rewritten, nearly from scratch
     - The Servlet API mocks are now standard Java classes, not `reify`-ed classes
     - A request body may now be a java.io.File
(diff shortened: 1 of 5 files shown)


hlship:
Note that this was backed out in 0.8.2-beta-4 (the :exceptions key is back).

PR #976, opened 2025-11-16
Jetty's context map must be representable.

souenzzo:
Currently if you capture jettys context map and try to inspect with some clojure data inspection tool, you will have problems.

It can't represent jetty response map because servlet-response `.toString` method throws an exception.

This PR creates a test that reproduces the issue
And proposes a solution.

I don't know if this is the ideal solution.
The best pratices about extending a multimethod says that to extend, you should own the class, or own the defmulti.


souenzzo:
https://github.com/pedestal/pedestal/actions/runs/19409304520/job/55528509431#step:9:523

PR #1014, opened 2026-06-10
chore: bump deps to latest

hlship:


embedded/resources/io/pedestal/embedded/build/deps.tmpl
```diff
@@ -34,6 +34,6 @@
   :build {:deps {io.github.clojure/tools.build {:mvn/version "0.10.13"}
```
souenzzo, line 34:
```suggestion
  :build {:deps {io.github.clojure/tools.build {:mvn/version "0.10.14"}
```

hlship: Good catch.  Keeping the embedded template up to date is currently far too manual. I need to write some scripts for this.

PR #1018, opened 2026-06-11
Release from CI on tag push

ericdallo:
Moves releases from manual local deploys to CI, triggered by pushing a version tag.

- New Release workflow: on a version tag push, verifies the tag matches `VERSION.txt`, builds all modules in dependency order and deploys them to Clojars, then creates a GitHub release with the changelog entry (beta/rc tags become prereleases)
- New `clj -T:build release` task: advances the version, stamps the changelog date, commits, tags and pushes; for final releases it also advances to the next patch SNAPSHOT and adds a fresh UNRELEASED changelog heading
- `deploy-all` no longer GPG signs by default, `:sign true` restores the old behavior
- See RELEASING.md for the full flow

Before the first release the repo needs the `CLOJARS_USERNAME` and `CLOJARS_PASSWORD` secrets, from a Clojars deploy token scoped to the io.pedestal group.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,78 @@
+name: Release
+on:
+  push:
+    tags:
+      - '[0-9]*'
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+      - name: Verify tag matches VERSION.txt
+        run: |
+          version="$(tr -d '[:space:]' < VERSION.txt)"
+          if [ "$version" != "$GITHUB_REF_NAME" ]; then
+            echo "Tag $GITHUB_REF_NAME does not match VERSION.txt ($version)"
+            exit 1
+          fi
+      - name: Setup Java
+        uses: actions/setup-java@v5
+        with:
+          java-version: '17'
+          distribution: 'corretto'
+      - name: Setup Clojure
+        uses: DeLaGuardo/setup-clojure@13.6.1
+        with:
+          cli: 1.12.5.1654
+
+      - name: Cache Clojure dependencies
+        uses: actions/cache@v5
+        with:
+          path: |
+            ~/.m2/repository
+            ~/.gitlibs
+            ~/.deps.clj
+          # List all files containing dependencies:
+          key: cljdeps-${{ hashFiles('deps.edn', '**/deps.edn') }}
+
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
+name: Release
+on:
+  push:
+    tags:
+      - '[0-9]*'
```
hlship, line 5: What if we want to release a beta?  Also, will this match `0.8.2` for example?

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+      - name: Verify tag matches VERSION.txt
```
hlship, line 12: Will this re-trigger when we push to master after a release?  Not sure what GITHUB_REF_NAME looks like in those cases.

build/build.clj
```diff
@@ -274,6 +286,97 @@
...
+    {:new-version  new-version
+     :final?       final?
+     :next-version (when final?
+                     (-> new-version
+                         v/parse-version
+                         (v/advance :patch)
+                         (v/advance :snapshot)
+                         v/unparse-version))}))
+
+(defn release
```
hlship, line 337: So this part is run manually?  I've been sloppy to date, tagging and pushing directly to master, but we probably should start creating a PR for releases and when those merge to master, will trigger the action, right?

.github/workflows/release.yml
ericdallo: yes it will match, this regex is basically a number and anything after that, so  all of these will work: `0.8.2`, `0.8.2-beta-9`, `1.0.0-rc-1`

.github/workflows/release.yml
ericdallo: no, this will trigger only for tags pushed, check `on: push: tags:`

build/build.clj
ericdallo: yes, we run `clj -T:build release` to trigger the release.
I dunno, never used a process of release with PRs because always is a maintainer doing it and there is nothing to review since it's just changelog that is already on master?

.github/workflows/release.yml
hlship: Oh, its a grep, not a regexp.

hlship (approved):

PR #1022, opened 2026-09-04
fix: mock servlet completed the response on the first flush

ericdallo:
The mock Servlet API behind `io.pedestal.test/response-for` counts a latch down when the response is complete, and since 6840c5b3 that happens on `ServletOutputStream.flush()` and `flushBuffer()`, besides `AsyncContext.complete()`. Flushing commits a response, it does not end it, and body writers flush freely: transit-java flushes about 30 times while writing a small map.

Two consequences:

- Async handler + a body written with several flushes: `response-for` returns on the first flush with a partial body. We hit this at Nubank with transit responses from async handlers (35 of 567 bytes, flaky tests).
- Sync handler whose body write throws before any flush (e.g. a closed `InputStream`): the servlet has returned but the latch never fires, so `response-for` waits for the whole timeout.

Also, `isCommitted()` has been returning false since that commit, because nothing set `responseCommitted` anymore.

This goes back to the 0.7 semantics: flushing only marks the response as committed; a synchronous response is complete when the servlet returns, an asynchronous one when the `AsyncContext` is completed. The `:timeout` option stays as a safety net for the async case.

Tests added for both cases in `http_test.clj`; both fail on master. Full suite passes.


tests/test/io/pedestal/http_test.clj
```diff
@@ -170,6 +197,21 @@
...
+  ;; flush of the output stream.
+  (let [response (response-for (app) :get "/async-chunked")]
+    (is (= 200 (:status response)))
+    (is (= "HELLO" (:body response)))))
+
+(deftest sync-response-is-complete-when-the-servlet-returns
+  ;; The body write fails before anything is flushed; a synchronous response is complete
+  ;; once the servlet returns, so this must not wait for a completion that never comes.
+  (let [response (response-for (app) :get "/failing-body" :timeout 1000)]
+    (is (= 200 (:status response)))
```
marcobiscaro2112: it took me a second to understand the expected 200, I think it's worth a comment about it (status + headers already sent to client, but body is empty because it threw)

marcobiscaro2112 (approved):
Looks good

